PR #36, opened 2015-03-26
Added Strain exercise

ErikSchierboom:
This PR implements the Strain exercise. In C#, the exercise is meant to work for all sequences and one is expected to write an extension method. However, I thought it would be nice for people to know that they can very easily extend the `Seq` module and thus add the new `keep` and `discard` methods into that module. 

However, as people may be unaware on how to do that, I have followed the approach taken in some Haskell exercises where a basic implementation file is automatically provided, to help the user use the correct structure for the code. An example of this can be found here: https://github.com/exercism/xhaskell/tree/master/zipper Is this something you are all comfortable with? Should I give make the basic implementation file compile, or not (I have opted for the latter)?

### DIFF
--- a/config.json
+++ b/config.json
@@ -28,7 +28,8 @@
     "allergies",
     "anagram",
     "clock",
-    "series"
+    "series",
+    "strain"
   ],
   "deprecated": [
 
--- a/strain/Example.fs
+++ b/strain/Example.fs
@@ -1,0 +1,5 @@
+﻿module Seq
+
+    let keep pred xs = seq { for x in xs do if pred x then yield x }
+
+    let discard pred = keep (not << pred)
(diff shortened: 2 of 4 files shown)


ErikSchierboom:
@jwood803 Do you have any idea why this PR fails to build on Travis? I can't seem to find anything wrong with the `config.json`, which it is complaining about.

kytrinyx:
Did you figure out the travis thing? I mean, you must have, since it looks to be passing, but I would love to improve the output of the `configlet` tool to be more helpful.

ErikSchierboom:
@kytrinyx Actually, I got some help with that. There was nothing wrong with the `configlet` tool, it said that there was an error in the `config.json` which there was, I just failed to notice a missing `"`. Nothing wrong with the tool, only with my eyes :(

kytrinyx:
I wonder if there's a simple command-line JSON linter that could say which line it's on. I don't think the go parser does that kind of granularity.

PR #68, opened 2016-03-15
Add queen-attack exercise

ErikSchierboom:


ErikSchierboom:
@jwood803 I've done some researching, and this PR is the first in what I propose to be the new exercise format. The most noteworthy changes:

- The tests are defined as `let` methods directly in a module, the type definition is not necessary.
- The test names are defined using double backticks, which allows test names to consists of easily readable strings containing spaces)
- The implementation is also done using `let` methods, unless the exercises requires an object-oriented approach. I try to use functional-first as the approach here.

What do you think?

jwood803:
Perfect! I love it! Been doing the same approach in my F# stuff lately.

ErikSchierboom:
@jwood803 Great! I'll merge it then.

PR #82, opened 2016-03-22
Add ocr-numbers exercise

ErikSchierboom:


exercises/ocr-numbers/Example.fs
```diff
@@ -0,0 +1,105 @@
...
+let matchCharacter input = 
+    match Map.tryFind input digitPatterns with
+    | Some x -> x
+    | None -> "?"
+
+let rows (input:string) = 
+    input.Split('\n')
+    |> Array.chunkBySize characterHeight
+
+let rowToCharacters (row: string[]) =
```
jwood803: I'm curious, will this work without having to specify what type it is and see if it can infer it?

ErikSchierboom: Yes this works :) The code works both locally in VS and remote on Travis.

jwood803: Sweet! Just something I noticed when messing with some of my stuff. The type inference is really good!

ErikSchierboom: Absolutely! Sometimes I find myself just trying things out, removing types, to see if it still compiles, and often it still does!

PR #136, opened 2016-06-04
Fix #135 (missing question mark)

### LINKED ISSUE #135: Missing question mark in one test
The Wordy test suite is missing a `?` in the "What is 4 minus -12" question. Since I had written a "is the question well-formed?" test into my solution that (among other things) tested for a question mark at the end, that meant that my code failed on that one unit test. My guess is that that difference wasn't intended, and that the "What is 4 minus -12" test should have been "What is 4 minus -12?" instead.

rmunn:
One question in the Wordy test suite was missing a question mark, making it inconsistent with the other tests.


ErikSchierboom:
Thanks!

PR #183, opened 2016-08-02
Add transpose exercise

ErikSchierboom:
This PR adds the [transpose](https://github.com/exercism/x-common/blob/master/transpose.md) exercise. 

@kytrinyx There is currently no test data specified for the transpose exercise, so this PR is my initial attempt. Do we also want to include some error checking? For example if there are rows with a different number of columns and vice versa? I'd be happy to convert submit a PR to x-common to add test data to the transpose exercise, if we can agree upon which test data we want.

### DIFF
--- a/config.json
+++ b/config.json
@@ -98,7 +98,8 @@
     "markdown",
     "all-your-base",
     "rectangles",
-    "sgf-parsing"
+    "sgf-parsing",
+    "transpose"
   ],
   "deprecated": [    
     "binary",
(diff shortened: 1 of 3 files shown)


kytrinyx:
> Do we also want to include some error checking?

That's an excellent question. How many exercises have error checking? I feel like since it's farther along in the track, it makes more sense to have error checking than if it were one of the earliest exercises... but if it seems like it detracts too much from the problem, I'm also completely happy to skip it.

Wow, that was perhaps the longest way of saying "I have no idea" that I've ever come up with.

ErikSchierboom:
I'll add some error checking :)

ErikSchierboom:
@kytrinyx In the end, I've decided to exclude the error checking, to make the exercise slightly easier.

PR #195, opened 2016-08-15
Update PangramTest.fs with another edge case

istepura:
Looking a solutions at (http://exercism.io/tracks/fsharp/exercises/pangram) I found one that would pass all the unit test and yet be invalid since it does not check whether symbols belong to alphabet 'a'..'z'


ErikSchierboom:
Excellent find! Thanks a lot for the help.

Would you mind also sending a PR to the [pangram.json](https://github.com/exercism/x-common/blob/master/pangram.json) file in the `x-common` repository? That way all tracks will benefit from this extra test.

ErikSchierboom:
Merged! 🎉 🎉 🎉

PR #199, opened 2016-08-26
Normalize test module names

robkeim:
After the normalization of the test files as *Test.fs there were some module names that weren't conforming to the *Test format.


ErikSchierboom:
Great work! Thanks. 🎉

PR #201, opened 2016-08-28
Align Result discriminated union terminology

robkeim:
The Result DU was defined as a mix of T/TSuccess and TError/TFailure. I aligned to TSuccess/TError based on the fact that F# is going to include this DU with those names in version 4.1.


ErikSchierboom:
Thanks! 🎉

PR #207, opened 2016-09-03
Fix all-your-base test. For great justice!

### LINKED ISSUE #206: Tests for all-your-base look wrong
I just got the new all-your-base exercise for the first time. And as I was looking through the unit tests to see the spec I'm implementing against, I noticed that an empty list `[]` is considered to be invalid input, rather than 0. That's a debatable issue (and I would disagree), but that's a place where opinion can validly differ -- if you have a string with no digits in it at all, it's fair enough to consider that to be "not a number" rather than representing 0, so a list with no digits could also represent 0. It will make my code slightly less elegant than it could have been, but fair enough.

But then I saw the next test, where the input is `[0]` and the output is `None`. That one's not debatable; it's just wrong. An input of a single 0 digit should produce the number 0 in the output base, and should not be an error.

And the next two tests after that, where the input is either multiple zeroes (`[0; 0; 0]`) or contains a leading zero (`[0; 6; 0]`), also expect an output of `None`. And here also, I believe the tests are wrong, and the correct answer should be 0, and "60" in whatever the input base is (in that particular test it's 7, so `60` is 42 in decimal).

Looking at https://github.com/exercism/x-common/blob/master/all-your-base.json for a bit, I figured out that the tests were automatically converted to F#, and the `null` results in that JSON file were turned into `None`. But the comment at the top of that JSON file says that it's those are tests where the right behavior could vary by language, and it's up to each language track to determine what the right behavior should be. I.e., they say that the correct representation of zero might be an empty list, or a list with a single zero, or a list with multiple zeroes... and that's why all those tests have a defined result of `null`. But at least *one* of them should be possible; as it stands, the F# version of the all-your-base exercise does not allow zero to be represented at all.

Which, in a test that was *named* after Zero Wing, just seems... well, **wrong** somehow.

The tests should be fixed... for great justice.

rmunn:
Fixes #206 by allowing any number of sequential zeroes to represent 0, but considering an empty list of digits to still be an error. (We can fix that one later if necessary).

### DIFF
--- a/exercises/all-your-base/AllYourBaseTest.fs
+++ b/exercises/all-your-base/AllYourBaseTest.fs
@@ -90,7 +90,7 @@ let ``Single zero`` () =
     let inputBase = 10
     let inputDigits = [0]
     let outputBase = 2
-    let outputDigits = None
+    let outputDigits = Some [0]
     Assert.That(rebase inputBase inputDigits outputBase, Is.EqualTo(outputDigits))
 
 [<Test>]
@@ -99,7 +99,7 @@ let ``Multiple zeros`` () =
     let inputBase = 10
     let inputDigits = [0; 0; 0]
     let outputBase = 2
-    let outputDigits = None
+    let outputDigits = Some [0]
     Assert.That(rebase inputBase inputDigits outputBase, Is.EqualTo(outputDigits))
 
 [<Test>]
@@ -108,7 +108,7 @@ let ``Leading zeros`` () =
     let inputBase = 7
     let inputDigits = [0; 6; 0]
     let outputBase = 10
-    let outputDigits = None
+    let outputDigits = Some [4; 2]
     Assert.That(rebase inputBase inputDigits outputBase, Is.EqualTo(outputDigits))
 
 [<Test>]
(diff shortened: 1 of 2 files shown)


rmunn:
The AppVeyor build runs the tests, but the Travis build doesn't? Weird.

... Oh, I see. The Travis build is honoring the `Ignore` attribute, but for some reason the AppVeyor build is actually *ignoring* the `Ignore` attribute and running the marked-as-ignored tests anyway. And the tests are failing because I updated the tests, but I didn't update the example code.

I'd rather not update the example code until after I do the exercise myself, so that I come at it fresh. If someone else wants to update the example, that would be great. Otherwise, I'll get to it, but it might be a couple of days.

ErikSchierboom:
Odd that the Travis build succeeds, it should have the same behavior. Both build servers should indeed run all tests, including those marked as ignored. The rationale behind that is that we want to be sure that our example code passes all tests. 

I can look into fixing the example code monday probably.

ErikSchierboom:
@rmunn I saw that you posted a correct implementation that works with the modified tests on exercism. Would you mind adding your correct implementation to this PR as the example implementation for this exercise?

rmunn:
@ErikSchierboom New example implementation added, all tests passing. PR should be ready to merge now.

ErikSchierboom:
Merged. Thanks! 🎉

PR #215, opened 2016-09-11
Remove Ignore attributes on tests for refactoring exercises

robkeim:
The refactoring exercises already have all of the tests passing so there's no need to progressively un-ignore them like when you're building a solution to an exercise.


ErikSchierboom:
Agreed :)

ErikSchierboom:
Thanks!

PR #224, opened 2016-10-03
Remove deprecated problems key

ErikSchierboom:


robkeim:
Nice to see this duplication get removed!

PR #235, opened 2016-12-08
Move `()` outside of unit test names so they are functions rather than values

theprash:
Just a small correction :)


ErikSchierboom:
Good find, thanks a lot! 🎉

PR #236, opened 2016-12-08
Fix tests to expect Option values

theprash:


robkeim:
Strange that this didn't pass the tests... @ErikSchierboom is there a way we can re-run the tests? We don't want to accept pull requests that aren't passing the tests, but the failure doesn't look related to these changes.

ErikSchierboom:
@robkeim Well apparently, AppVeyor changed something in their configuration. The part that breaks this script will be [this line](https://github.com/exercism/xfsharp/blob/master/build.fsx#L44). I've created an [issue](#237) for it.

In general, the Travis build is the more important of the two, as it also checks the validity of the config file. Unfortunately, there is also a [Travis bug](#209), which causes the tests to not run. 

I cannot run the tests again, as @kytrinyx is the owner of the repository and thus the AppVeyor account linked to this repository.

@theprash The PR looks great, but would you mind if I looked at it tomorrow? I should really try to fix the build issues first.

ErikSchierboom:
@theprash I've just merged this [PR](https://github.com/exercism/xfsharp/pull/238) to fix the AppVeyor build. In the process, I've also used the changes you suggested, as otherwise the build failed and I couldn't be certain I fixed the bug. Would you mind rebasing to see if there is anything left to merge? Thanks a lot for finding this by the way.

robkeim:
Thanks @ErikSchierboom for the explanation/fix and echoing Erik's point thanks for the contribution @theprash!

theprash:
No problem. Your changes are fine. Thanks!

ErikSchierboom:
@theprash Thanks you once again!

Issue #234, opened 2016-12-08
Add hints to some exercises


ErikSchierboom:
@abo64 added [this issue](https://github.com/exercism/xscala/issues/220) to the Scala track, which I thought was a brilliant plan. I'm just gonna literally copy-and-paste that issue's text:

>Most exercises have their own favorable solutions and characteristic pitfalls.
To give people some guidance and avoid writing repetitive or similar code comments again and again
we should add a HINTS.md file to any exercise where it makes sense.

>Especially the first couple of exercises should have such a file to assist beginners.
Another case is exercises dealing with special features of the language like `option` or `async`.

We should make a list of exercises for which some hints can be added. Of course, the first exercises should be more hint-heavy than the latter, although even for advanced exercises hints can be very useful. One example of this are the text parsing exercises, where we could guide people to using FParsec.

ErikSchierboom:
The following exercises could probably use some hints:

- [ ] `helloworld`: give hints on [pattern matching](https://fsharpforfunandprofit.com/posts/match-expression/) and [sprintf](https://fsharpforfunandprofit.com/posts/printf/)
- [ ] `difference-of-squares` give hint on `Seq.sumBy`
- [ ] `sum-of-multiples`: give hint on using ranges
- [ ] `space-age`: give hint that code duplication should be avoided and that `match` expressions are favored over dictionaries
- [ ] `accumulate`: give hints about making a [tail-call recursive function](https://blogs.msdn.microsoft.com/fsharpteam/2011/07/08/tail-calls-in-f/)
- [ ] `grains`: give hint on `BigInt`
- [ ] `wordy`: give hint on using [FParsec](http://www.quanttec.com/fparsec/users-guide/)
- [ ] `diamond`: give hint on [property-based testing](https://fsharpforfunandprofit.com/posts/property-based-testing/)

TODO: the rest

robkeim:
Great idea @abo64 and thanks @ErikSchierboom for bringing this to the F# track as well. I've gone ahead and sent a first exercise [here](https://github.com/exercism/xfsharp/pull/241) and if it looks good I'll try to take care of some of the others as well.

ErikSchierboom:
@robkeim I've updated the list with more suggested hints.

robkeim:
Thanks @ErikSchierboom you're making fixing this too easy by going as far as to include links :) I sent a PR with a few additional ones, and I'll try to get some more done soon.

ErikSchierboom:
Fantastic!

[...]

PR #241, opened 2016-12-11
Add hints for grains test (#234)

robkeim:
This is to address #234 (adding hints for exercises)

@ErikSchierboom This was inspired by the work you did on the Scala track [here](https://github.com/exercism/xscala/pull/222/files). Is this what you had in mind? If so, then I'll try to add some more as well.


ErikSchierboom:
This is *precisely* what I had in mind! Thanks a ton. 🎉 

I'll work on updating the list of exercises that I feel could use a hint.

robkeim:
Great, thanks for your precise descriptions!

PR #247, opened 2016-12-13
Add another good tail recursion reference

ErikSchierboom:
Although the existing tail recursion link is great, [this post](http://blog.ploeh.dk/2015/12/22/tail-recurse/) by ploeh is also a great explanation. Is it worth it to also include this in the hints?


robkeim:
This is a great article, thanks for the good read and the PR @ErikSchierboom!

Issue #240, opened 2016-12-10
Linked List exercise tests imply an immutable list


theprash:
I think the tests imply that the doubly linked list is expected to be an immutable data structure because:

* All functions return a new version of the list.
* The tests do not check for mutation of the list passed in.
* Most importantly, `mkLinkedList` is a plain value, not a function, so the empty list at least must be immutable to avoid individual tests affecting each other.

Although immutability is generally preferred in F#, it seems like it would be too difficult to create an immutable doubly linked list for an exercise like this. So I suggest changing the tests so that:

* The functions don't return the new list.
* Tests check for mutation of the list that was passed in.
* `mkLinkedList` is a function that takes unit.

This would make the requirements clearer and also simplify the tests, which would no longer need lines like this:

```fsharp
let (val''''''', linkedList''''''') = pop linkedList''''''
```

ErikSchierboom:
Well, we had this discussion a while ago. Back then, we decided that we would try to make as many exercises use immutable data as possible, which is after all the FP way. However, I can see what you're getting at. The test do look quite strange. What do you feel @robkeim and @jwood803?

theprash:
I can see the reason for preferring immutability, but there is a time and place for mutation. Currently, this exercise seems like a confusing mix of both, with a description that's gone out of sync with the tests. Almost all of the submitted answers use mutation and even example answer does!

I can see how the tests are almost set up in a way that would allow you to use either approach. That could be a valid way to go, but to be completely flexible, `mkLinkedList` should be a function.

ErikSchierboom:
I can totally see what you're getting at. I would just like to get the opinion of someone else. @rmunn @robkeim @jwood803 ?

robkeim:
I'm definitely for keeping immutability as it's the FP where it makes sense, however, I definitely see the point @theprash is making and this case looks a bit "forced."

I tried to be very dogmatic throughout my own implementation of these exercises only using immutable data, but I caved for this one. A doubly-linked list isn't naturally a FP structure and I'm sure the creation of this exercise was done without FP in mind.

It looks like no one that has solved this problem so far has done it without using mutable state, so I'm not against modifying this to use mutable state. That change should also be coupled with an explicit call out in the description that this exercise doesn't follow FP's immutable style.

rmunn:
[...]

PR #248, opened 2016-12-14
Change linked-lists tests and example to use mutation.

theprash:
For #240.

I also fixed an error in `pop` and `shift` where next/prev elements weren't being set to `None` and the list was keeping references to items that had been removed. The tests do not check for this.

### DIFF
--- a/exercises/linked-list/Example.fs
+++ b/exercises/linked-list/Example.fs
@@ -1,36 +1,43 @@
 module LinkedList
 
 type Element<'a> = { value: 'a; mutable prev: Element<'a> option; mutable next: Element<'a> option }
-type LinkedList<'a> = { first: Element<'a> option; last: Element<'a> option }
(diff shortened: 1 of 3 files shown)


ErikSchierboom:
This looks great! Thanks.

I have one small request: could you add a `HINTS.md` file to the exercise's directory? In this file we should link to a page that helps users understand how to use classes in F#. I think [this page](https://fsharpforfunandprofit.com/posts/classes/) would be a great resource to include. You can see an example of how such an `HINTS.md` file looks like [here](https://github.com/exercism/xfsharp/pull/241/files).

theprash:
Why classes? The example implementation still just uses records with mutable fields.

theprash:
I've added hints mentioning both records and classes.

ErikSchierboom:
Oh sorry, I should have of course mentioned records too. I think the hints file is perfect. Merged!

Thanks a lot for all the work! 🎉

robkeim:
Nice job @theprash these changes look good!

PR #249, opened 2016-12-15
Prevent permission issues in grep exercise

ErikSchierboom:
To prevent permission issues, we write the file in the user's temp directory.


robkeim:

PR #252, opened 2016-12-17
Fix incorrect tree building tests and implementation. Fixes #250

### LINKED ISSUE #250: Tree Building exercise in F# has nonsense readme
From the readme:

> The ID number is always
between 0 (inclusive) and the length of the record list (exclusive). No record
has a parent ID lower than its own ID and no record, except the root record,
has a parent ID that's equal to its own ID.

These sentences are both false and violated in the tests. Maybe the design changed and someone forgot to update the readme? Currently the root node has a parent id of -1.

The root node, and at least one child of the root, must have "a parent ID lower than its own ID".

ErikSchierboom:


ErikSchierboom:
Note: I've merged the PR myself due to the fact that this was a bug that needed to be fixed ASAP.

exercises/tree-building/Example.fs
```diff
@@ -5,13 +5,16 @@ type Tree =
...
+
 let addOrAppend key value map =
     let list = defaultArg (Map.tryFind key map) []
     Map.add key (list @ [value]) map
 
 let invalidNode previous x = 
-    x.ParentId >= x.RecordId || 
-    x.RecordId <> previous + 1
+    match x.RecordId with
+    | 0 -> x.ParentId <> rootNodeRecordId
```
robkeim: `| rootNodeRecordId -> x.ParentId <> rootNodeRecordId`

robkeim: Can you re-use the variable here you defined above?

robkeim:

exercises/tree-building/Example.fs
ErikSchierboom: Nope, because that's not how F# pattern matching works. It will not match on the value, but on the variable with that name. Therefore, it would then match on any input. For an example, see [this fiddle](https://dotnetfiddle.net/GA3oIn).

robkeim: Interesting, thanks for the explanation @ErikSchierboom... you learn something new everyday :)

PR #257, opened 2016-12-18
Add HINTS to difference-of-squares exercise

ErikSchierboom:
As specified in #234.

### DIFF
--- a/exercises/difference-of-squares/HINTS.md
+++ b/exercises/difference-of-squares/HINTS.md
@@ -1,3 +1,3 @@
 ## Hints
 For this exercise the following F# feature comes in handy:
-- [Seq.sumBy](https://msdn.microsoft.com/en-us/visualfsharpdocs/conceptual/seq.sumby%5B't,%5Eu%5D-function-%5Bfsharp%5D) is a condensed format to apply a function to a sequence and then sum the results
+- The [range operator](https://msdn.microsoft.com/en-us/visualfsharpdocs/conceptual/operators.%5B-..-%5D%5B%5Et%5D-function-%5Bfsharp%5D) allows you to succinctly create a range of values.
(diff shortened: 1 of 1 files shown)


exercises/difference-of-squares/HINTS.md
```diff
@@ -1,3 +1,3 @@
 ## Hints
 For this exercise the following F# feature comes in handy:
-- [Seq.sumBy](https://msdn.microsoft.com/en-us/visualfsharpdocs/conceptual/seq.sumby%5B't,%5Eu%5D-function-%5Bfsharp%5D) is a condensed format to apply a function to a sequence and then sum the results
```
robkeim: @ErikSchierboom is there a reason that you don't think the `Seq.sumBy` hint is useful for this exercise?

Most people end up creating a square function and then doing something like this:
```
let sumOfSquares num =
    [1..num]
    |> List.sumBy square
```

The reason I had added this was I remember learning about the `sumBy` after having done it in two different lines.

exercises/difference-of-squares/HINTS.md
```diff
@@ -1,3 +1,4 @@
 ## Hints
 For this exercise the following F# feature comes in handy:
```
robkeim: @ErikSchierboom now there are two items in the list it needs to be "features come" instead of "feature comes" :)

ErikSchierboom: Good find! I've updated the PR. By the way, if you merge this, can you do that using the "Squash and merge" option? In general, this is how I merge PR's, as we then end up with one combined commit, instead of several smaller ones. cc @jwood803

robkeim: Sounds good, thanks @ErikSchierboom. I didn't know what our merging policy was, but I'll stick to "squash and merge" in the future!

robkeim:
Thanks @ErikSchierboom for adding this hint!

PR #258, opened 2016-12-18
Make Markdown exercise consistent by never expecting `<p>` inside `<li>`

theprash:
The tests here expect to see a `<p>` inside `<li>` only when the text isn't contained in some other tag such as `<em>`. See these snippets from two different tests:

```fsharp
    let expected = "<ul><li><p>Item 1</p></li><li><p>Item 2</p></li></ul>"
    ...
    let expected = "<h1>Header!</h1><ul><li><em>Bold Item</em></li><li><i>Italic Item</i>
```

Neither GitHub markdown nor [CommonMark](http://spec.commonmark.org/dingus/) behave like this so I couldn't see a reason to expect it.

I've removed the expected `<p>` in the tests and updated both implementations. I intentionally left behind the unused variable `notusep` in the "bad" implementation.

### DIFF
--- a/exercises/markdown/Example.fs
+++ b/exercises/markdown/Example.fs
@@ -1,6 +1,5 @@
 ﻿module Markdown
 
-open System
 open System.Text.RegularExpressions
 
 let openingTag tag = sprintf "<%s>" tag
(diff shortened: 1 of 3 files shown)


robkeim:
@theprash thanks for the PR! The changes look good to me, but I'm going to let @ErikSchierboom review/merge this since he authored the original exercise and might have more context on the current `<p>` behavior.

theprash:
Here are some more points of reference:

* The [Elixir Markdown exercise](https://github.com/exercism/xelixir/blob/a327ebce7b51fdd3bc4cb468295510487ea9c150/exercises/markdown/markdown_test.exs) - Does not have `<p>` in `<li>`

* The [Markdown renderer for exercism.io itself](https://github.com/exercism/exercism.io/blob/3de39152058110ace5b1ea98f3f9915d13480294/test/exercism/markdown_test.rb) - Does not have `<p>` in `<li>`

* The [original Markdown syntax doc](https://daringfireball.net/projects/markdown/syntax#list) - Does not have `<p>` in `<li>`

* The [PHP Markdown exercise](https://github.com/exercism/xphp/blob/34fd62906a947f8139292fe1eac9cc2a2d0b623f/exercises/markdown/markdown_test.php) - **Does** have `<p>` in `<li>`

It seems like a mistake that started off in one of the language tracks' tests and has proliferated a bit.

ErikSchierboom:
@theprash This is great! Very good find, well explained and fixed! Thanks a lot. 🎉

PR #261, opened 2016-12-20
Add hints to palindrome-products exercise

ErikSchierboom:


robkeim:
@ErikSchierboom I'm not sure I understand what your hint is trying to suggest here. What are the slow and more efficient methods you are referring to?

When I did this exercise I just converted the number to a string, reversed the string, then tested if they were equal to let .NET do the heavy lifting for me. I did that because it was the first thing that came to mind and I didn't take any computational complexity into concern, but don't remember that being problematic.

ErikSchierboom:
I also tried to use a string reversal for the palindrome test, but for the large numbers this was quite slow. I then managed to speed this up significantly using the following algorithm:

```fsharp
let isPalindrome n = 
    let mutable current = n
    let mutable result = 0
    while(current > 0) do
        result <- result * 10 + current % 10
        current <- current / 10

    result = n
```

But you didn't experience any particular 'slowless' for the string reversal check?

robkeim:
Ok, I just double checked and found this in my tests file for this problem:
`// This test is commented out due to performance reasons (it takes ~1 min to run)`

So you're totally right my solution does have performance problems :) I like the solution you've got!

ErikSchierboom:
I figured as much :)

PR #262, opened 2016-12-20
Add hints for nth-prime exercise

ErikSchierboom:


robkeim:
@ErikSchierboom I assume you're referring to the [Sieve of Eratosthenes](https://en.wikipedia.org/wiki/Sieve_of_Eratosthenes) here? It might be helpful to also add a link in case the text doesn't help. What do you think?

ErikSchierboom:
Well yes and no. The idea of the Sieve of Erastosthenes is what I'm hinting at, but I would not suggest to users that they use it to implement this exercise. The main reason is that with the Sieve of Erastosthenes you have to know beforehand how many numbers you'll check, which is hard to determine if you want to find the 10001st prime. To solve this exercise more efficiently, you should use lazy, infinite sequences. Maybe I should explain this also in the hint?

robkeim:
@ErikSchierboom yes I think it would make more sense to be more explicit in the hints to be consistent with the other ones. Here I would call out infinite sequences using the "The following feature is useful for this exercise" format.

ErikSchierboom:
@robkeim The PR has been updated.

robkeim:
Thanks @ErikSchierboom for the changes and not doing a rebase and leaving the two commits so I could review the diff :)

I've gone ahead and _squashed_ and committed your changes!

PR #279, opened 2017-01-12
Use correct ignore string in Luhn test

ErikSchierboom:
When using TestCase, the Ignore string must be exactly equal to "Remove to run test case". Otherwise, the CI servers don't run the test.

### DIFF
--- a/exercises/luhn/LuhnTest.fs
+++ b/exercises/luhn/LuhnTest.fs
@@ -4,10 +4,10 @@ open NUnit.Framework
 open Luhn
   
 [<TestCase("1", ExpectedResult = false)>] // single digit strings can not be valid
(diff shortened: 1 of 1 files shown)


robkeim:
@ErikSchierboom I assume the C# track has the same issue then since I made the same [change](https://github.com/exercism/xcsharp/commit/dac6c863e0dcccf44c76caca3e1dc2bb796289ad) in both tracks. Are you going to do the change there as well?

ErikSchierboom:
Yes I will, later :)

robkeim:
Sounds good, thanks for fixing this for me @ErikSchierboom!

ErikSchierboom:
You're welcome! You can run the test script locally too by the way, by running `build.cmd` or `build.sh`. That's how I found out :)

robkeim:
Good to know!

PR #280, opened 2017-01-12
Refactored tree building tests. Fixes #250

### LINKED ISSUE #250: Tree Building exercise in F# has nonsense readme

ErikSchierboom:
This PR makes the tree building tests not force any specific implementation upon the user. For the discussion, see #250.

### DIFF
--- a/exercises/tree-building/Example.fs
+++ b/exercises/tree-building/Example.fs
@@ -5,6 +5,12 @@ type Tree =
     | Branch of int * Tree list
     | Leaf of int
 
+let recordId = function Branch (id, _) | Leaf id -> id
+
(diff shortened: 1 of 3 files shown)


robkeim:

PR #283, opened 2017-01-16
Remove test from triangle. Closes #282

### LINKED ISSUE #282: Triangle - Test case Triangles violating triangle inequality are illegal 2 is incorrect
The test case asserts that a triangle with sides 2 2 4 violates the Triangle inequality theorem; however, it is a valid Isosceles triangle.

ErikSchierboom:
As discussed in #282, the 2 4 2 triangle test is marked as incorrectly marked as invalid. The correct solution according to the [canonical data](https://github.com/exercism/x-common/blob/master/exercises/triangle/canonical-data.json) is to remove this test:
> we have decided NOT to test triangles where all side lengths are positive but a + b = c., e.g: (2, 4, 2, Isosceles), (1, 3, 4, Scalene).",

This PR removes said test.

### DIFF
--- a/exercises/triangle/TriangleTest.fs
+++ b/exercises/triangle/TriangleTest.fs
@@ -70,10 +70,5 @@ let ``Triangles violating triangle inequality are illegal`` () =
 
 [<Test>]
 [<Ignore("Remove to run test")>]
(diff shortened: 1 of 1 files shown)


robkeim:
Nice work @ErikSchierboom, thanks!

PR #287, opened 2017-01-22
Simple Cipher: add missing Ignore attribute

balazsbotond:
The `Ignore` attribute was missing from the test case `Cipher can double shift encode` so I added it.


ErikSchierboom:
Great! Thanks a lot. :tada

PR #288, opened 2017-01-22
Simple cipher: fix comment grammar

balazsbotond:
The original comment said: "Here we take advantage of the fact that plaintext of "aaa..." doesn't output the key." If I understand correctly, this is the opposite of what was meant, that is, "Here we take advantage of the fact that plaintext of "aaa..." outputs the key."


ErikSchierboom:
More thanks!

PR #289, opened 2017-01-22
Fix #286 - Linked List: change pop and shift to return option

### LINKED ISSUE #286: Linked List: how to handle the empty list in pop and shift?
There is currently no unit test checking what `pop` and `shift` do when the list is empty. I wonder if returning an `option` would be a better way to handle this case instead of returning the value if present or throwing an exception. Though `Map` has separate `find` and `tryFind` methods so I'm not sure if this is idiomatic.

I think adding tests for the empty case would be a good idea in any case and if you agree, I can write them. And if you think `option` is the way to go, I will change that too.

balazsbotond:


robkeim:
I agree this is more idiomatic, but from the [description.md](https://github.com/exercism/x-common/blob/master/exercises/linked-list/description.md) it says:
> To keep your implementation simple, the tests will not cover error conditions. Specifically: pop or shift will never be called on an empty list.

Shouldn't we stay aligned with what's in x-common?

PR #293, opened 2017-01-30
Add hint about encoding structure in code to song exercises

ErikSchierboom:
This PR adds a hint to the song-like exercise, stating that people should try to encoding the structure of the song. The beer-song exercise already had this hint.


robkeim:
Nice work, thanks @ErikSchierboom!

PR #313, opened 2017-02-21
Added test cases to enforce correct interaction between matching flags

nicholaslawton:
The test suite for the Grep exercise does not enforce correct behaviour for interaction between all combinations of flags.

This pull request adds four new test cases to verify the following combinations:
 - case insensitive and entire line matching
 - case insensitive matching and inversion
 - inversion and entire line matching
 - all three

See [issue 310](https://github.com/exercism/xfsharp/issues/310).


ErikSchierboom:
Merged! :tada:

PR #316, opened 2017-02-21
Modify test cases using foldr - issue 314

hobbitjoe:
modified the test cases that use foldr so that the syntax reflects the same order of the accumulator and foldable list that is used in List.foldBack


ErikSchierboom:
Looks great! Thanks a lot for reporting the issue and fixing it! :tada:

PR #337, opened 2017-07-31
Mark exercises as core or unlocked_by

ErikSchierboom:


robkeim:
@ErikSchierboom is there any easy way to visualize the tree of dependencies for the exercises to make this easier to review or did you just construct the dependency tree mentally? I was thinking something like this:

```
A
B
+--- G
     +--- H
     +--- I
C
+--- F
E
```

C unlocks F, B unlocks G which in turn unlocks H and I.

ErikSchierboom:
@robkeim Yes, you can upload the config.json file on [this website](https://jonmcalder.shinyapps.io/exercism-config-viz/) to create a visualization.

robkeim:
That visualization is really really @ErikSchierboom and way more detailed than I was envisioning :)

Great work on this, I like the progression but there are three exercises that aren't part of the core track and not dependent on anything else, is that intentional? If so, what's the reasoning for having the floaters?

ErikSchierboom:
@robkeim Well, I didn't make the visualization, so I can't take credit :) The rationale behind the "floating" problems can be found [here](https://github.com/exercism/discussions/issues/175).

There is a small bug in this PR where one problem's entry (`rectangles`) is missing. I'll fix that soon.

ErikSchierboom:
The `rectangles` bug has been fixed by the way, so if you are okay with the current setup, feel free to merge.

robkeim:
Great work again @ErikSchierboom!

ErikSchierboom:
@robkeim Thanks for reviewing!

PR #340, opened 2017-08-17
Add two-fer exercise

robkeim:
I unfortunately can't run the build script locally anymore with fake so if there are any build errors I'll fix them.

### DIFF
--- a/config.json
+++ b/config.json
@@ -46,6 +46,17 @@
         "Transforming"
       ]
     },
+	{
+      "uuid": "2ee3cc7a-db3f-4668-9983-ed6d0fea95d1",
+      "slug": "two-fer",
+      "core": true,
+      "unlocked_by": "null",
+      "difficulty": 1,
+      "topics": [
+        "Strings",
+        "Optional values"
+      ]
+    },
     {
       "uuid": "277d05db-0ba0-4de6-b5f8-090c251afffc",
       "slug": "space-age",
--- a/config.json
+++ b/config.json
@@ -192,8 +192,8 @@
     {
       "uuid": "4e786e56-2658-445f-ac91-64dd9c38dbb3",
       "slug": "binary-search-tree",
-      "core": true,
-      "unlocked_by": null,
+      "core": false,
+      "unlocked_by": "two-fer",
       "difficulty": 5,
       "topics": [
         "Searching",
@@ -780,7 +780,7 @@
       "uuid": "cb629d30-0351-4023-bd51-423267164c24",
       "slug": "bank-account",
       "core": false,
-      "unlocked_by": "binary-search-tree",
+      "unlocked_by": "two-fer",
       "difficulty": 5,
       "topics": [
         "Optional values",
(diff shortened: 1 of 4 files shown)


robkeim:
I definitely need to invest some time into making the build script work because trial and error isn't going to scale, particularly when I'm making the changes directly on github.com on the metro :)

config.json
```diff
@@ -46,6 +46,17 @@
         "Transforming"
       ]
     },
+	{
+      "uuid": "2ee3cc7a-db3f-4668-9983-ed6d0fea95d1",
+      "slug": "two-fer",
+      "core": true,
```
ErikSchierboom: I agree this should be a core exercise, but we can probably then remove the `binary-search-tree` as a core exercise as its topics will already have been introduces by earlier exercises. 

We should also have the `two-fer` exercise unlock a couple of non-core exercises, which might require some reshuffling due to the possible removal of the `binary-search-tree` as a core exercise, which will then no longer unlock any exercises.

exercises/two-fer/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+Note that, when trying to submit an exercise, make sure the exercise file that you're submitting is in the `exercism/csharp/<exerciseName>` directory.
+
+For example, if you're submitting `bob.cs` for the Bob exercise, the submit command would be something like `exercism submit <path_to_exercism_dir>/csharp/bob/bob.cs`.
+
+## Source
+This is an exercise to introduce users to basic programming constructs, just after Hello World. [https://en.wikipedia.org/wiki/Two-fer](https://en.wikipedia.org/wiki/Two-fer)
+
+## Submitting Incomplete Solutions
+It's possible to submit an incomplete solution so you can see how others have completed the exercise.
```
ErikSchierboom: Should we be adding a hints to documentation on optional values? E.g. [this page](https://fsharpforfunandprofit.com/posts/the-option-type/).

ErikSchierboom:
> I definitely need to invest some time into making the build script work because trial and error isn't going to scale, particularly when I'm making the changes directly on github.com on the metro :)

Yes, the build script is not the best of things. However, I wouldn't put too much effort in it currently, as with the release of .NET Core 2.0, we could start using .NET Core for the track, with stub files, project files, etc. Unfortunately, the tooling in VS 2017 doesn't correctly work with F# on .NET Core 2.0, so we have to wait for a tooling update to work on that.

Thanks for doing this by the way!

config.json
robkeim: @ErikSchierboom which exercises did you think this should unlock? I got the inspiration for the fact that this was a core exercise from the C# track, but didn't think to look at what exercises (if any) it should unlock.

exercises/two-fer/README.md
robkeim: Definitely, this was a shameless copy/paste from the C# track :)

ErikSchierboom: :)

config.json
ErikSchierboom: @robkeim Well actually, I have to update the C# track as it doesn't unlock any exercises there ATM. The way I decided which exercises to unlock, is to check which 5/6 exercises have one or more matching topics and were not unlocked by a previous core exercise.

robkeim:
@ErikSchierboom I've updated the PR with your comments

Regarding the build, I agree we should be moving to .NET core, I just forgot how much work you've put into the C# track and now this one feels like it's in the dark ages :)

ErikSchierboom:
Great work! Thanks! 🎉

PR #347, opened 2017-08-22
Add default introductory code example

kytrinyx:
Instead of having logic for a fallback in the backend, we're choosing a
default for the snippet file.

For tracks that have core exercises, we pick the first core exercise.
For tracks without these, we pick the first exercise listed in the config.

See https://github.com/exercism/meta/issues/89

### DIFF
--- a/docs/SNIPPET.txt
+++ b/docs/SNIPPET.txt
@@ -1,0 +1,3 @@
+﻿module HelloWorld
+
+let hello = "Hello, World!"


robkeim:
Thanks @kytrinyx! 🎉 

@ErikSchierboom regarding this snippit do you think there is something else that we should have instead? If we want to go with the most simple example like we did for the C# track perhaps we should remove the module declaration and just have a one liner, although then we're not really showing off much of the language.

ErikSchierboom:
@robkeim I had the same thought, so I wanted to let the module declaration in just to show off a bit of the language.

Issue #309, opened 2017-02-17
Create project for each exercise


ErikSchierboom:
To help users get started programming, we should provide the user with a skeleton .fsproj file. This .fsproj file should work with the .NET CLI and run on .NET Core.

hobbitjoe:
@ErikSchierboom some coworkers and I have been using visual studio code along with a make file using Fake in order to build the code and run the tests without the need of a fsproj file.  Since we do not use a fsproj file it looks like the test file has errors highlighted incorrectly.  Not sure if you would be interested in something like this or not, but figured I'd let you know how I've been running the f# tests.

ErikSchierboom:
@hobbitjoe Cool. The [build script](https://github.com/exercism/xfsharp/blob/master/build.fsx) does something similar. With the .NET CLI tooling (and updated .fsproj files) almost being finished, I think we won't be needing such a FAKE script (hopefully). Would you mind if I contact you to test things once I'll be working on this issue?

hobbitjoe:
@ErikSchierboom sure, I'll be happy to help test!

ErikSchierboom:
Great! I'll keep you to that promise :)

ErikSchierboom:
Fixed by #348. @hobbitjoe Would you mind testing it (by fetching a new exercise) to see if it works?

hobbitjoe:
@ErikSchierboom sorry for the delay but I was able to fetch a new problem today and test it out.  I was able to run the tests via the CLI using dotnet core 2.0 on macOS 10.12.6 with no issues.  Let me know if any other information is needed.

ErikSchierboom:
@hobbitjoe Great! Thanks for testing.

Issue #308, opened 2017-02-17
Allow exercises to build and run on .NET Core


ErikSchierboom:
This issue will keep track of the work items to get our exercises running on .NET Core.

- [x] Convert tests from NUnit to xUnit (#297)
- [x] Create a project for each exercise (#309)
- [x] Fix the CI builds
- [x] Update the documentation

Once these items have been completed, we probably want to to wait at least until the final release of Visual Studio 2017 (which will be on march 7th) and the RTM version of the .NET CLI tooling. In the mean time, I would really really appreciate it if people would be willing to look at this PR and hopefully also test things locally.

To make testing and editing all projects at the same time easier, I've also added a solution file that contains all projects.

When this PR has finally been merged, we can then move on to issue #229, where we add a default file for each exercise, possibly also with the required functions also having an empty, stub implementation.

ErikSchierboom:
Fixed by #348.

Issue #298, opened 2017-02-07
Replace [<TestCase>] tests with individual tests


ErikSchierboom:
We have several places where we use `[TestCase]` to create a single test with multiple input values. While efficient (no code duplication), this is not the best of user experiences as it makes it harder to run a single test and also doesn't really allow us to mark the various input values with a descriptive text. It also doesn't align with how canonical data for exercises is defined, which only works on single test cases.

Consider the `isogram` exercises's tests:

```fsharp
[<TestCase("duplicates", ExpectedResult = true)>]
[<TestCase("eleven", ExpectedResult = false, Ignore = "Remove to run test case")>]
[<TestCase("subdermatoglyphic", ExpectedResult = true, Ignore = "Remove to run test case")>]
[<TestCase("Alphabet", ExpectedResult = false, Ignore = "Remove to run test case")>]
[<TestCase("thumbscrew-japingly", ExpectedResult = true, Ignore = "Remove to run test case")>]
[<TestCase("Hjelmqvist-Gryb-Zock-Pfund-Wax", ExpectedResult = true, Ignore = "Remove to run test case")>]
[<TestCase("Heizölrückstoßabdämpfung", ExpectedResult = true, Ignore = "Remove to run test case")>]
[<TestCase("the quick brown fox", ExpectedResult = false, Ignore = "Remove to run test case")>]
[<TestCase("Emily Jung Schwartzkopf", ExpectedResult = true, Ignore = "Remove to run test case")>]
[<TestCase("éléphant", ExpectedResult = false, Ignore = "Remove to run test case")>]
let ``Isogram correctly detects isograms`` (actual: string) =
    isogram actual
```

What's the difference between test case one and two? It's not immediately obvious. The [canonical-data](https://github.com/exercism/x-common/blob/master/exercises/isogram/canonical-data.json) of this exercise looks much nicer. I think we should convert the [<TestCase>] tests to individual tests.

robkeim:
This is also being discussed in the C# track https://github.com/exercism/xcsharp/issues/194

robkeim:
Here are the list of exercises using the `TestCase` attribute:
- [ ] [acronym](https://github.com/exercism/xfsharp/tree/master/exercises/acronym)
- [ ] [atbash-cipher](https://github.com/exercism/xfsharp/tree/master/exercises/atbash-cipher)
- [ ] [clock](https://github.com/exercism/xfsharp/tree/master/exercises/clock)
[...]

Issue #297, opened 2017-02-07
Replace NUnit with Xunit


ErikSchierboom:
Xunit has slowly but steadily been replacing NUnit as the favored testing framework. Virtually all open-source Microsoft code now uses Xunit as the unit-testing framework. The integration with .NET Core is thus also better, as Microsoft has invested in Xunit for its own software and thus has to keep it running. An example of this improved integration is the fact that there is an Xunit template available in the .NET CLI.

Note: we should first do #203, as that would make this transition hopefully just a matter of changing package references :)

robkeim:
The same conversion is happening in C# as part of https://github.com/exercism/xcsharp/issues/193

ErikSchierboom:
Fixed by #348.

PR #348, opened 2017-08-29
Support .NET Core

### LINKED ISSUE #203: Using FsUnit for more expressive tests
At the moment, the exercises used NUnit's `Assert` class' methods to verify behavior. While this works, it's not really a nice functional API. Perhaps it would make sense to use [FsUnit](http://fsprojects.github.io/FsUnit/), which provides a functional wrapper around the tests.

As an example, this is how an assertion looks like today:

``` fsharp
Assert.That(1, Is.EqualTo(1))
```

And this is how it would look with FsUnit:

``` fsharp
1 |> should equal 1
```

I much prefer the second option. It would require people to install an additional package before the tests can be run, but that won't really be a problem as I'm working on creating a pre-defined project for each exercise, which would already contain the required libraries.

cc @jwood803 @rmunn @robkeim @rebelwarrior @pminten

ErikSchierboom:
This PR fixes several issues:

- Create a project for each exercise. This makes getting started a lot simpler. (#309)
- All the exercises now have a default, stub file. This also makes getting started easier. All the easy (difficulty 1-3) exercises have stub methods with fully specified types, also to help people get started. The medium (difficulty 4-7) exercises have stub methods without types (type-inference) and the hard exercises (difficulty 8-10) only have an empty stub module. (#229)
- Allow exercises to run on .NET Core. This is great for our cross-platform support. Users on Mac or Unix should now have the same options (and command-line interface) as Windows users. (#308)
- FsUnit is used to make the tests appear more functional. (#203)

What is left to do is:
- Replace `[<TestCase>]` tests with individual tests. (#298)
- Replace NUnit with Xunit. This should be easier now that we've moved to FsUnit. (#297)
- Fix Travis build.

Note: it is still a work-in-progress, but if you'd be willing to take a look, that would be great!

CC: @jpreese Would you also mind taking a look?

This fixes #203, #229, #251, #277, #297, #308, #309


build.cake
```diff
@@ -0,0 +1,91 @@
...
+                System.IO.File.WriteAllText(testFile.FullPath, updatedContents);
+            }
+        }
+    });
+
+Task("TestRefactoringProjects")
+    .IsDependentOn("EnableAllTests")
+    .Does(() => {
+        var refactoringSolution = ParseSolution(refactoringSolutionPath);
+        Parallel.ForEach(refactoringSolution.Projects, parallelOptions, (project) => DotNetCoreTest(project.Path.FullPath));
```
robkeim: Do you need the `noBuild` configuration setting like in the C# track?
https://github.com/exercism/csharp/pull/387/files

exercises/Exercises.All.sln
```diff
@@ -0,0 +1,1502 @@
...
+Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "Wordy", "wordy\Wordy.fsproj", "{B14FF411-3EDE-4E2F-98AB-78CBDFFBAF48}"
+EndProject
+Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "ZebraPuzzle", "zebra-puzzle\ZebraPuzzle.fsproj", "{3C57692F-7E65-4C6F-9F2D-3962A9007EE2}"
+EndProject
+Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "Zipper", "zipper\Zipper.fsproj", "{1DEE9AFA-D88A-4143-B53D-E494FF331FC8}"
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
+		Debug|x64 = Debug|x64
```
robkeim: Minor: we don't need `x64` or `x86` builds the `Any CPU` is enough

exercises/acronym/AcronymTest.fs
```diff
@@ -1,64 +1,65 @@
 module AcronymTest
 
 open NUnit.Framework
+open FsUnit
 open Acronym
 
 [<Test>]
 let ``Empty string abbreviated to empty string`` () =
-    Assert.That(acronym "", Is.EqualTo(""))
+    acronym "" |> should equal ""
```
robkeim: `should be NullOrEmptyString`

generate.fsx
```diff
@@ -0,0 +1,118 @@
+open System
```
robkeim: Should this be checked in? It seems like a tool that you are using for the migration but it won't be useful after that.

ErikSchierboom: It should indeed be removed. I was planning to wait for this, but as I don't really need it anymore, I'll remove it now.

exercises/acronym/AcronymTest.fs
ErikSchierboom: Well, I don't think it should actually. The test case states explicitly that it should be abbreviated to an empty string. If I were to use `NullOrEmptyString`, returning `null` would also pass the test.

exercises/Exercises.All.sln
ErikSchierboom: Good point!

build.cake
ErikSchierboom: No I don't, as the .NET CLI version 2.0 will automatically restore packages when the `dotnet test` command runs. In the C# track, we use the older version of the CLI which requires an explicit `dotnet restore` before that.

exercises/allergies/AllergiesTest.fs
```diff
@@ -1,74 +1,79 @@
...
-    Assert.That(allergicTo Allergen.Strawberries 0, Is.False)
+    allergicTo Allergen.Peanuts 0 |> should equal false
+    allergicTo Allergen.Cats 0 |> should equal false
+    allergicTo Allergen.Strawberries 0 |> should equal false
 
 [<Test>]
 [<Ignore("Remove to run test")>]
 let ``Allergic to eggs`` () =
-    Assert.That(allergicTo Allergen.Eggs 1, Is.True)
+    allergicTo Allergen.Eggs 1 |> should equal true
```
ErikSchierboom: @robkeim What do you think about using `should equal true`? The docs also mention `should be True`, but I was hesistant to use that option as the uppercase T in `True` could throw newcomers off I think,

exercises/acronym/AcronymTest.fs
ErikSchierboom: Note: there also is a `EmptyString` variant, which I'll use.

build.cake
robkeim: Ah nice!

exercises/acronym/AcronymTest.fs
robkeim: Sounds good!

generate.fsx
robkeim: Ok!

exercises/allergies/AllergiesTest.fs
robkeim: Given there's already a mix of `should be` and `should equal` I'm fine going with `should equal true`.

ErikSchierboom:
@robkeim The PR has been updated and I think it is now correctly working. Could you perhaps review the PR again? If you deem it correct, feel free to merge.

docs/TESTS.md
```diff
@@ -1,34 +1,21 @@
...
-Click on "New Project" and you should see an entry for Visual F# and click on "Class Library".
+```bash
+dotnet test
+```
 
-![New Project](http://x.exercism.io/v3/tracks/fsharp/docs/img/newProject.png)
+## Solving the exercise
 
-Install [NUnit](http://nunit.org/index.php?p=download) via NuGet package manager with the NUnit.Console package. This will add the reference to the project.
+Solving an exercise means making all its tests pass. By default, only one test (the first one) is executed when you run the tests. This is intentional, as it allows you to focus on just making that one test pass. Once it passes, you can enable the next test by removing `Skip = "Remove to run test"` from the test's `[<Fact>]` or `[<Theory?]` attribute. When all tests have been enabled and your implementation makes them all pass, you'll have solved the exercise!
```
robkeim: Minor typo: `[<Theory>]`

exercises/acronym/AcronymTest.fs
robkeim: @ErikSchierboom did you decide not to update this to `EmptyString` for some reason?

docs/TESTS.md
robkeim: Fixed!

ErikSchierboom:
@robkeim I forgot about the EmptyString. Perhaps I can fix it later in a new commit?

robkeim:
@ErikSchierboom yes I figured you could do it in another PR. Once the fix of my typo is green I'll go ahead and merge this branch.

robkeim:
Thanks for your hard work on this @ErikSchierboom! Great job 🎉 🎉 🎉

ErikSchierboom:
Hurray!

ErikSchierboom:
image

@Mousaka @kytrinyx The F# track has now been updated to support .NET Core, which is its biggest overhaul yet!

kytrinyx:
@ErikSchierboom Fantastic. Thanks so much for taking the time to do this overhaul... this is huge.

PR #350, opened 2017-09-03
Allow building single exercise through build script

ErikSchierboom:
You can do this by adding: `--name=<exercise>`

This makes it easier to test new exercises or changes to existing ones.


robkeim:
Useful feature, thanks @ErikSchierboom!

Issue #353, opened 2017-09-24
TreeBuilding exercise should not have implemented functions


danielhensilva:
I've fetched TreeBuilding and it seems to already be implemented with all tests cases passing. 
Also, there are some missing `skip` for them.

ErikSchierboom:
@danielhensilva This is not actually a bug, but a feature! The tree building exercise is an unusual exercise in that it starts with a working implementation. Its goal is to refactor the implementation into something better, which is also why none of the test cases are being skipped (as you can verify the refactoring didn't break anything).

I hope this clears things up!

danielhensilva:
Oh! I've got it now! Sorry about this, Erik! :)

ErikSchierboom:
No problem!

PR #354, opened 2017-09-24
Remove implemented functions (#353)

danielhensilva:
And add missing skip props for test cases.


ErikSchierboom:
Not actually an issue, see my comment on #535.

PR #357, opened 2017-10-07
Fix type of roster in stub

### LINKED ISSUE #356: grade-school: Map<int, string list> compared with (int * string list) list
In the exercises stub, there is the function `roster` stubbed:

https://github.com/exercism/fsharp/blob/master/exercises/grade-school/GradeSchool.fs#L7

```
let roster (school: Map<int, string list>): Map<int, string list> = failwith "You need to implement this function."
```

In a later test its return value is compared to a list of key-value-pairs:

https://github.com/exercism/fsharp/blob/master/exercises/grade-school/GradeSchoolTest.fs#L52-L66

```
let ``Student names and grades in roster are sorted`` () =
    let school =
        empty        
        |> add "Jennifer" 4
        |> add "Kareem" 6
        |> add "Christopher" 4
        |> add "Kyle" 3

    let expected = 
        [(3, ["Kyle"]);
         (4, ["Christopher"; "Jennifer"]);
         (6, ["Kareem"])]

    roster school |> should equal expected
```

In the example implementation said function is implemented as `let roster school = Map.toList school`, so it has a list type infered.

NobbZ:
FIX #356

### DIFF
--- a/exercises/grade-school/GradeSchool.fs
+++ b/exercises/grade-school/GradeSchool.fs
@@ -4,6 +4,6 @@ let empty: Map<int, string list> = failwith "You need to implement this function
 
 let add (student: string) (grade: int) (school: Map<int, string list>): Map<int, string list> = failwith "You need to implement this function."
 
-let roster (school: Map<int, string list>): Map<int, string list> = failwith "You need to implement this function."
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
Good catch! Thanks. 🎉

PR #359, opened 2017-10-07
Really enforce sorted student names in roster

### LINKED ISSUE #358: grade-school: ``Student names and grades in roster are sorted`` does miss edge-cases
When items are appended to the internal list, the test works as expected, but I do prepend students, therefore that test misses my case and passes, without me ever sorting the roster…

```fsharp
let add (student: string) (grade: int) (school: Map<int, string list>): Map<int, string list> =
    school
    |> Map.tryFind grade
    |> (fun students -> defaultArg students [])
    |> (fun students -> student::students)
    |> (fun students -> school.Add(grade, students))

let roster (school: Map<int, string list>): (int * string list) list =
    Map.toList school
```

NobbZ:
By adding three names in neither ascending, nor descending alphabetical order, we can guarantee that the roster has been sorted.

The former way of testing could miss implementations that `cons`ed instead of `append`.

FIX #358

### DIFF
--- a/exercises/grade-school/GradeSchoolTest.fs
+++ b/exercises/grade-school/GradeSchoolTest.fs
@@ -57,10 +57,11 @@ let ``Student names and grades in roster are sorted`` () =
         |> add "Kareem" 6
         |> add "Christopher" 4
         |> add "Kyle" 3
+        |> add "Zoe" 4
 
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
Thanks again!

PR #360, opened 2017-10-08
Generators

ErikSchierboom:
This PR adds basic generators support for the F# track. It is quite a big PR, so take your time reviewing it.


generators/Exercises.fs
```diff
@@ -0,0 +1,448 @@
...
+        let value = this.RenderValueWithoutIdentifier (canonicalDataCase, key, value)
+        sprintf "let %s = %s" identifier value  
+
+    default this.RenderIdentifier (canonicalDataCase, key, value) = String.camelize key
+
+    default this.RenderIdentifierWithTypeAnnotation (canonicalDataCase, key, value) = 
+        let identifier = this.RenderIdentifier (canonicalDataCase, key, value)
+    
+        match this.IdentifierTypeAnnotation (canonicalDataCase, key, value) with
+        | Some ty -> identifier |> addTypeAnnotation ty
```
robkeim: Why `ty`?

generators/Exercises.fs
```diff
@@ -0,0 +1,448 @@
...
+
+    default this.PropertiesWithIdentifier canonicalDataCase = []
+
+    default this.IdentifierTypeAnnotation (canonicalDataCase, key, value ) = None
+
+    default this.UseFullMethodName canonicalDataCase = false
+
+    default this.AdditionalNamespaces = []
+
+type Acronym() =
```
robkeim: What's the reason for having all of the exercises in the same file as opposed to each one in a separate file?

generators/Exercises.fs
```diff
@@ -0,0 +1,448 @@
...
+        | "create", "expected", _ -> value :?> int64 <> -1L |> box
+        | _ -> base.MapCanonicalDataCaseProperty (canonicalDataCase, key, value)
+
+    override this.RenderInput (canonicalDataCase, key, value) =
+        let parsePositionTuple (tupleValue: obj) =
+            let parts = 
+                (tupleValue :?> JToken)
+                    .SelectToken("position")
+                    .ToObject<string>()
+                    .TrimStart('(')
```
robkeim: `Trim('(')` should trim both the start and end

generators/Options.fs
```diff
@@ -0,0 +1,31 @@
...
+
+open System
+open System.IO
+open CommandLine
+open Serilog
+open Humanizer
+
+type Options = {
+  [<CommandLine.Option('e', "exercises", Required = false, 
+    HelpText = "Exercises to generate (if not specified, defaults to all exercises).")>] Exercises : seq<string>;
```
robkeim: For the `HelpText` it might be helpful to have example values for each of these values. For example, it's not clear to be how to specify multiple exercises to be generated here (I'm assuming some sort of delimiter but I'd have to try whitespace, semicolon, etc). For the canonical data directory, can it be a relative pass as well as absolute?

generators/Output.fs
```diff
@@ -0,0 +1,224 @@
...
+    let indentation = String.replicate level "    "
+    sprintf "%s%s" indentation value
+
+let parenthesize value = sprintf "(%s)" value
+
+let backwardPipe value = sprintf "<| %s" value
+
+let backwardPipeConditional test value = if test value then backwardPipe value else value
+
+let addTypeAnnotation ty value = sprintf "%s: %s" value ty
```
robkeim: I'm still wondering what `ty` is :)

generators/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "profiles": {
+    "Generators": {
+      "commandName": "Project",
+      "commandLineArgs": "-e allergies"
```
robkeim: Should this line be checked in? Isn't this only for testing?

ErikSchierboom: It is. I'll remove it.

generators/Output.fs
ErikSchierboom: It's the type (e.g. `"int list'`). I'll rename.

generators/Exercises.fs
ErikSchierboom: Because `type` is a reserverd keyword. But I'll rename, you're right.

generators/Exercises.fs
ErikSchierboom: Yes, but I don't want that. I want to trim the start '(' and trim the end ')'.

generators/Exercises.fs
ErikSchierboom: As far as I can tell, having them all in one file is more idiomatic than have each one in a separate file. In general, as far as I can tell, each file corresponds to a single module, which may include many classes.

However, it may be better to split the `Exercise` class from its implementations. I'll try it to see how it works out.

generators/Options.fs
ErikSchierboom: @robkeim I think perhaps these things can be explained in a README (which I plan to add later), as they require a little bit more explanation than I expect of a help text.

ErikSchierboom:
@robkeim I've fixed your PR comments, thanks for reviewing!

generators/Exercises.fs
robkeim: Ah ok that makes sense :)

generators/Exercises.fs
robkeim: That's really surprising that it's idiomatic because I feel like having them all in the same file goes against isolation. We don't have a lot of people concurrently working on these generators normally, but I feel like if we did having them in the same file would lead to more merge conflicts.

generators/Options.fs
robkeim: Sounds good @ErikSchierboom!

generators/Exercises.fs
ErikSchierboom: @robkeim Well, as in F# you have to specify the dependencies of files in the right order, having lots of small files becomes a bit more tedious than in C#. Here are some examples I found:

- https://github.com/fsharp/FAKE/blob/master/src/app/Fake.Core.Target/Target.fs
- https://github.com/fsharp/FSharp.Compiler.Service/blob/master/src/fsharp/CompileOps.fs
- https://github.com/fsprojects/FSharpLint/blob/master/src/FSharpLint.Core/Framework/Ast.fs
- https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/Common/Domain.fs

I see your point about merge conflicts, but as long as not too many people are working on the same exercise, we'll probably be fine I guess.

The next step is to create individual issues for the test generators I think, just like we did for the C# track.

robkeim: It's been too long since I've done F# I'm definitely getting rusty, thanks for the clarification @ErikSchierboom!

PR #408, opened 2017-10-16
Adding prime factors tests generator #395 .

vrnithinkumar:
Since the name in the exercise is "primeFactorsFor" canonical-data.json has "factors"
overrided RenderSutProperty with "primeFactorsFor".
issue #395

### DIFF
--- a/generators/Generators.fs
+++ b/generators/Generators.fs
@@ -170,6 +170,14 @@ type Pangram() =
 type PigLatin() =
     inherit Exercise()
 
+type PrimeFactors() =
+    inherit Exercise()
+    
+    override this.RenderInput (canonicalDataCase, key, value) =
+        base.RenderInput (canonicalDataCase, key, value) |> sprintf "%sL"
+    
+    override this.RenderSutProperty canonicalDataCase = "primeFactorsFor"
+
 type QueenAttack() =
     inherit Exercise()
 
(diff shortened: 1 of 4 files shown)


generators/Generators.fs
```diff
@@ -170,6 +170,14 @@ type Pangram() =
 type PigLatin() =
     inherit Exercise()
 
+type PrimeFactors() =
+    inherit Exercise()
+    
+    override this.RenderInput (canonicalDataCase, key, value) =
+        base.RenderInput (canonicalDataCase, key, value) |> sprintf "%sL"
+    
+    override this.RenderSutProperty canonicalDataCase = "primeFactorsFor"
```
ErikSchierboom: I would rather rename the property in our example and stub file. Could you remove this line and adjust the example and stub implementation to follow the new name?

vrnithinkumar: I will remove this line and adjust the stub implementation and examples.

Regarding the example implementation, I think you are right. As you said I tried building and running the implementation. Apparently it was running for a while and memory was shooting up.
Should we go change that too ?

ErikSchierboom: @vrnithinkumar If you would, yes please! It is important that our example implementations run in a reasonable time frame, as that directly influences the build server time.

ErikSchierboom:
@vrnithinkumar Looking great! I'll wait for the CI to become green, and then I'll merge.

ErikSchierboom:
Merged! Thanks a lot! :tada:

PR #410, opened 2017-10-18
Test case generation for perfect-numbers.

vrnithinkumar:
Added test case generation for perfect-numbers fIx for #391.
Added a InvalidInput in classification for invalid input data. 
Updated the example to support invalid input case.

### DIFF
--- a/exercises/perfect-numbers/Example.fs
+++ b/exercises/perfect-numbers/Example.fs
@@ -1,14 +1,14 @@
 ﻿module PerfectNumbers
 
-type Classification = Perfect | Abundant | Deficient | InvalidInput
+type Classification = Perfect | Abundant | Deficient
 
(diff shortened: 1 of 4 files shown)


robkeim:
@ErikSchierboom I agree with what you said and would also be in favor of option 3.

vrnithinkumar:
Option 3 seems like more apt to me. I will make those changes. @ErikSchierboom Please have a look.

ErikSchierboom:
The changes look good! The CI build failed for weird reasons, so I've just restarted the build. Once it is green, I'll merge.

ErikSchierboom:
There is something wrong with Travis. Will look into

ErikSchierboom:
And the build is back! Merged. Thanks a lot. :tada:

PR #412, opened 2017-10-20
Adding test generation got pascals-triangle

vrnithinkumar:
Adding test generation got pascals-triangle, issue #390.
In this case we have to handle 3 types of expected outputs, which are Empty, valid and invalid input case. So experimented with exception rather than using option type.

- I tried introducing a new template for test throwing exception in the case of invalid input. 
- In the example we are raising ArgumentOutOfRangeException in case of invalid input.
- Overriding `ToTestMethodBodyAssertTemplate` to identify the exception throwing cases and choose  template for exception.
- Kept the name of method as "triangle" rather than changing it to "row".

I am not sure is this the right way to tackle the issue. @ErikSchierboom , Could you please share your thoughts?

### DIFF
--- a/exercises/pascals-triangle/PascalsTriangleTest.fs
+++ b/exercises/pascals-triangle/PascalsTriangleTest.fs
@@ -29,5 +29,5 @@ let ``Four rows`` () =
 
 [<Fact(Skip = "Remove to run test")>]
 let ``Negative rows`` () =
-    triangle -1 |> should equal None
+    triangle -1 |> should throw typeof<ArgumentOutOfRangeException>
 
--- a/exercises/pascals-triangle/PascalsTriangleTest.fs
+++ b/exercises/pascals-triangle/PascalsTriangleTest.fs
@@ -9,25 +9,25 @@ open PascalsTriangle
 
 [<Fact>]
 let ``Zero rows`` () =
-    triangle 0 |> should be Empty
+    rows 0 |> should equal (Some ([]: int list list))
 
 [<Fact(Skip = "Remove to run test")>]
 let ``Single row`` () =
-    triangle 1 |> should equal [[1]]
+    rows 1 |> should equal (Some [[1]])
 
 [<Fact(Skip = "Remove to run test")>]
 let ``Two rows`` () =
-    triangle 2 |> should equal [[1]; [1; 1]]
+    rows 2 |> should equal (Some [[1]; [1; 1]])
 
 [<Fact(Skip = "Remove to run test")>]
 let ``Three rows`` () =
-    triangle 3 |> should equal [[1]; [1; 1]; [1; 2; 1]]
+    rows 3 |> should equal (Some [[1]; [1; 1]; [1; 2; 1]])
 
 [<Fact(Skip = "Remove to run test")>]
 let ``Four rows`` () =
-    triangle 4 |> should equal [[1]; [1; 1]; [1; 2; 1]; [1; 3; 3; 1]]
+    rows 4 |> should equal (Some [[1]; [1; 1]; [1; 2; 1]; [1; 3; 3; 1]])
 
 [<Fact(Skip = "Remove to run test")>]
 let ``Negative rows`` () =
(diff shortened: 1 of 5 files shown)


ErikSchierboom:
> I tried introducing a new template for test throwing exception in the case of invalid input.

Looks good!

> In the example we are raising ArgumentOutOfRangeException in case of invalid input.

I'm not sure if we should be using exceptions, as they are less idiomatic than using a Result or Option value. What do you think @robkeim? 

> Overriding ToTestMethodBodyAssertTemplate to identify the exception throwing cases and choose template for exception.

Perfect.

> Kept the name of method as "triangle" rather than changing it to "row".

In the [canonical data](https://github.com/exercism/problem-specifications/edit/master/exercises/pascals-triangle/canonical-data.json), the property is named "rows" (plural). I don't really mind using "rows", what about you @robkeim ?

generators/Generators.fs
```diff
@@ -197,6 +197,25 @@ type PerfectNumbers() =
         | "deficient" -> "(Some Deficient)"
         | _ -> "None"
 
+type PascalsTriangle() =
+    inherit Exercise()
+
+    override this.RenderSutProperty canonicalDataCase = "triangle"
```
ErikSchierboom: Could you remove this to just use the default property name?

exercises/pascals-triangle/PascalsTriangleTest.fs
```diff
@@ -22,9 +28,6 @@ let ``Four rows`` () =
     triangle 4 |> should equal [[1]; [1; 1]; [1; 2; 1]; [1; 3; 3; 1]]
 
 [<Fact(Skip = "Remove to run test")>]
-let ``Five rows`` () =
-    triangle 5 |> should equal [[1]; [1; 1]; [1; 2; 1]; [1; 3; 3; 1]; [1; 4; 6; 4; 1]]
+let ``Negative rows`` () =
+    (fun () -> triangle -1 |> ignore) |> should throw typeof<System.ArgumentOutOfRangeException>
```
ErikSchierboom: I would like to encode an error as `None`.

generators/Generators.fs
vrnithinkumar: Changed to default property name "rows" and parameter name to "numberOfRows"

exercises/pascals-triangle/PascalsTriangleTest.fs
vrnithinkumar: Changed it to use `int list list option` to return the result and `None` for error cases. 
- Made changes from `should be Empty` to `should equal (Some ([]: int list list))` for empty test case. 
https://stackoverflow.com/questions/23989847/fsunit-should-equal-fails-on-some 
- Overriding ToTestMethodBodyAssertTemplate to use only Assert Template.

generators/Generators.fs
```diff
@@ -200,21 +200,21 @@ type PerfectNumbers() =
...
+                |> formatList
+                |> sprintf "(Some %s)"
+
+        | _ -> "None"
 
-    override this.ToTestMethodBodyAssertTemplate canonicalDataCase =
-         match canonicalDataCase.Expected with
-         | :? JArray -> base.ToTestMethodBodyAssertTemplate canonicalDataCase
-         | _ -> "AssertThrows"
+    override this.ToTestMethodBodyAssertTemplate canonicalDataCase = "AssertEqual"
```
ErikSchierboom: Is this line necessary?

vrnithinkumar: Added that to skip using AssertEmpty template. In case of `Some [] |> should be Empty`, it's  failing.

ErikSchierboom: Ah that makes sense.

ErikSchierboom:
Thanks!

robkeim:
This is great work @vrnithinkumar! I agree with @ErikSchierboom regarding returning an `Option` type instead of throwing exceptions and `rows` sounds good to me :)

PR #413, opened 2017-10-20
Largest series product

vrnithinkumar:
Adding test case generator for largest series product . Fix for #382.
- Added "digit" as a properties with Identifier since it is very long some times.
- Added property "comment" in excluded list.
- Fixed example to handle invalid cases.

### DIFF
--- a/exercises/largest-series-product/LargestSeriesProductTest.fs
+++ b/exercises/largest-series-product/LargestSeriesProductTest.fs
@@ -1,40 +1,84 @@
+// This file was auto-generated based on version 1.0.0 of the canonical data.
+
 module LargestSeriesProductTest
 
-open Xunit
 open FsUnit.Xunit
-open System
+open Xunit
+
 open LargestSeriesProduct
-    
-[<Theory(Skip = "Remove to run test")>]
-[<InlineData("01234567890", 2, 72)>]
-[<InlineData("1027839564", 3, 270)>]
-let ``Gets the largest product``(digits: string) (seriesLength: int) (expected: int) =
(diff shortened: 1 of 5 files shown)


exercises/largest-series-product/LargestSeriesProductTest.fs
```diff
@@ -1,40 +1,84 @@
...
+
+[<Fact(Skip = "Remove to run test")>]
+let ``Reports 1 for nonempty string and empty product (0 span)`` () =
+    let digits = "123"
+    largestProduct digits 0 |> should equal 1
+
+[<Fact(Skip = "Remove to run test")>]
+let ``Rejects empty string and nonzero span`` () =
+    let digits = ""
+    largestProduct digits 1 |> should equal -1
```
ErikSchierboom: I would like to encode an error as `None`.

generators/Generators.fs
```diff
@@ -154,6 +154,11 @@ type KindergartenGarden() =
 
     override this.UseFullMethodName canonicalDataCase = true
 
+type LargestSeriesProduct() =
+    inherit Exercise()
+
+     override this.PropertiesWithIdentifier canonicalDataCase = ["digits"]
```
ErikSchierboom: Good idea to have this as an identifier!

exercises/largest-series-product/LargestSeriesProductTest.fs
vrnithinkumar: Made changes to test use `int option` as return type and use `None` for error case.

ErikSchierboom:
Merged. Thanks!

PR #415, opened 2017-10-23
Simplify generators

ErikSchierboom:
This PR does a couple of things:

1. It updates the kindergarten-garden generator and implementation to the updated canonical data.
1. It simplifies identifier handling by providing a method to retrieve all data-related properties.
1. The order of the identifiers is such that the expected value is always last (if present).
1. It simplifies formatting option values and make its usage consistent over all generators.
1. Option values are now parenthesized instead of using the backward pipe (`<|`) character, as the latter is harder to understand.

Additional note: don't squash this PR, as the three commits should be retained IMHO.


robkeim:
Thanks 🎉

PR #416, opened 2017-10-23
Adding test case generator for dominoes.

vrnithinkumar:
Adding test case generator for dominoes. #374
- Added a method to format the list as tuple
- Added "input" as PropertiesWithIdentifier since the input array is longer some times.

### DIFF
--- a/generators/Generators.fs
+++ b/generators/Generators.fs
@@ -100,6 +100,21 @@ type CryptoSquare() =
 type DifferenceOfSquares() =
     inherit Exercise()
 
+type Dominoes() =
+    inherit Exercise()
+    
+    let formatAsTuple (value:obj) =
+        let twoElementList = value :?> JArray |> normalizeJArray
+        (twoElementList.Item 0, twoElementList.Item 1) |> string
+
+    override this.RenderInput (canonicalDataCase, key, value) =
+        value :?> JArray
+        |> normalizeJArray
+        |> Seq.map formatAsTuple
+        |> formatList
+
+    override this.PropertiesWithIdentifier canonicalDataCase = ["input"]
+
 type Gigasecond() =
     inherit Exercise()
 
@@ -147,7 +162,7 @@ type KindergartenGarden() =
 type LargestSeriesProduct() =
     inherit Exercise()
 
-     override this.PropertiesWithIdentifier canonicalDataCase = ["digits"]
+    override this.PropertiesWithIdentifier canonicalDataCase = ["digits"]
 
     override this.RenderExpected (canonicalDataCase, key, value) = 
         value 
--- a/generators/Generators.fs
+++ b/generators/Generators.fs
@@ -113,7 +113,7 @@ type Dominoes() =
         |> Seq.map formatAsTuple
         |> formatList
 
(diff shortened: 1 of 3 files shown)


generators/Generators.fs
```diff
@@ -100,6 +100,21 @@ type CryptoSquare() =
...
+        let twoElementList = value :?> JArray |> normalizeJArray
+        (twoElementList.Item 0, twoElementList.Item 1) |> string
+
+    override this.RenderInput (canonicalDataCase, key, value) =
+        value :?> JArray
+        |> normalizeJArray
+        |> Seq.map formatAsTuple
+        |> formatList
+
+    override this.PropertiesWithIdentifier canonicalDataCase = ["input"]
```
ErikSchierboom: With #415 merged, you can now also use `this.Properties canonicalDataCase` (see [example](https://github.com/exercism/fsharp/pull/415/files#diff-378e45c2fa29f82b16b9d312d5f5c18aR173)) instead of being explicit when defining the properties with an identifier. The benefit of that is that renamed values don't break the generator, although it also means that new properties will automatically have an identifier. In this case I'm partial to either solution, what do you think?

generators/Generators.fs
```diff
@@ -147,7 +162,7 @@ type KindergartenGarden() =
 type LargestSeriesProduct() =
     inherit Exercise()
 
-     override this.PropertiesWithIdentifier canonicalDataCase = ["digits"]
+    override this.PropertiesWithIdentifier canonicalDataCase = ["digits"]
```
ErikSchierboom: See comment above

generators/Generators.fs
vrnithinkumar: I agree with the idea of using `this.Properties canonicalDataCase` rather than defining the properties with an identifier. Will change to use  `this.Properties canonicalDataCase`.
Thanks @ErikSchierboom

generators/Generators.fs
vrnithinkumar: Will change to use this.Properties canonicalDataCase.

generators/Generators.fs
ErikSchierboom: @vrnithinkumar Ah, I linked to the wrong property, as `this.Properties` also includes the expected value, which doesn't need to be included. Could you use `this.PropertiesUsedAsSutParameter` instead? Sorry about that.

generators/Generators.fs
ErikSchierboom: @vrnithinkumar Ah, I linked to the wrong property, as `this.Properties` also includes the expected value, which doesn't need to be included. Could you use `this.PropertiesUsedAsSutParameter` instead? Sorry about that.

vrnithinkumar: Made changes to use  Using PropertiesUsedAsSutParameter not Properties

generators/Generators.fs
vrnithinkumar: Made changes to use  Using PropertiesUsedAsSutParameter not Properties

ErikSchierboom:
Looking good! Just waiting for CI to become green.

robkeim:
Changes look good to me and CI is green, thanks @vrnithinkumar 🎉

PR #417, opened 2017-10-24
Implementation for Robot Simulator Test Generator

roman-shuhov:
Added Robot Simulator test generator to fix issue #399

### DIFF
--- a/generators/Generators.fs
+++ b/generators/Generators.fs
@@ -267,6 +267,96 @@ type RailFenceCipher() =
 type Raindrops() =
     inherit Exercise()
 
+type RobotSimulator() =
+    inherit Exercise()
+
+    member this.ResultIdentifierName = "actual"
+
+    override this.PropertiesWithIdentifier canonicalDataCase = ["robot"; "property"; "expected"]
+
+    member private this.RenderDirection (value: JToken) = 
+        sprintf "%s" (value.ToObject<string>() |> String.upperCaseFirst)  
+
+    member private this.RenderCoords (coords: JToken) = 
+        (coords.["x"].ToObject<int>(), coords.["y"].ToObject<int>()) 
+        |> formatValue
+
+    member private this.DefineRobot (direction: JToken) (coords: JToken) = 
+        sprintf "createRobot %s %s" (this.RenderDirection direction) (this.RenderCoords coords)
+
+    member private this.GetRobotProperties (value : JToken) = 
+        value.SelectToken("direction"), value.SelectToken("position")
+
+    override this.RenderArrange canonicalDataCase =
+        // one identifier may be empty if we only checking created object
+        // we need to filter out this empty line
+        base.RenderArrange canonicalDataCase
+        |> List.choose (
+            function
+            | "" -> None
+            | v -> Some v
+        )
+
+    override this.RenderValueWithoutIdentifier (canonicalDataCase, key, value) = 
+        match key with
+        | "robot" ->
+            let input = value :?> JToken
+            this.DefineRobot ("direction" |> input.SelectToken) ("position" |> input.SelectToken)
+        | "expected" -> 
+            // here we may need to render full robot object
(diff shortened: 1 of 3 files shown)


generators/Generators.fs
```diff
@@ -267,6 +267,96 @@ type RailFenceCipher() =
 type Raindrops() =
     inherit Exercise()
 
+type RobotSimulator() =
+    inherit Exercise()
+
+    member this.ResultIdentifierName = "actual"
```
ErikSchierboom: Why not use a simple identifer: `let resultIdentifierName = "actual"`?

generators/Generators.fs
```diff
@@ -267,6 +267,96 @@ type RailFenceCipher() =
...
+        match key with
+        | "robot" ->
+            let input = value :?> JToken
+            this.DefineRobot ("direction" |> input.SelectToken) ("position" |> input.SelectToken)
+        | "expected" -> 
+            // here we may need to render full robot object
+            // or just coordinate/direction values
+            let input = (value :?> JToken)
+            let direction, position = this.GetRobotProperties input
+            match isNull(direction), isNull(position) with
```
ErikSchierboom: Do you need the parentheses? Or could `isNull(direction)` be `isNull direction`? The same applies to the second `isNull` call.

generators/Generators.fs
```diff
@@ -267,6 +267,96 @@ type RailFenceCipher() =
...
+
+    override this.RenderValueWithoutIdentifier (canonicalDataCase, key, value) = 
+        match key with
+        | "robot" ->
+            let input = value :?> JToken
+            this.DefineRobot ("direction" |> input.SelectToken) ("position" |> input.SelectToken)
+        | "expected" -> 
+            // here we may need to render full robot object
+            // or just coordinate/direction values
+            let input = (value :?> JToken)
```
ErikSchierboom: You can lose the parentheses

generators/Generators.fs
```diff
@@ -267,6 +267,96 @@ type RailFenceCipher() =
...
+            | false, false ->
+                this.DefineRobot ("direction" |> input.SelectToken) ("position" |> input.SelectToken)
+            | true, true -> 
+                ""
+        | _ -> ""
+
+    override this.RenderValueWithIdentifier (canonicalDataCase, key, value) = 
+        match key with 
+        | "property" -> 
+            let action = (value :?> string)
```
ErikSchierboom: You can lose the parentheses

generators/Generators.fs
```diff
@@ -267,6 +267,96 @@ type RailFenceCipher() =
...
+            | true, true -> 
+                this.ResultIdentifierName
+            | false, false -> 
+                match canonicalDataCase.Property with
+                | "create" -> "robot"
+                | _ -> this.ResultIdentifierName
+    
+    override this.RenderTestMethodName canonicalDataCase =
+        // avoid duplicated method names
+        sprintf "%s - %s" canonicalDataCase.Property (canonicalDataCase.Description |> String.upperCaseFirst)
```
ErikSchierboom: There is another option to avoid duplicated method names, see [this example](https://github.com/exercism/fsharp/blob/master/generators/Generators.fs#L145). However, I've looked at the canonical data, and it looks like that would lead to incredibly long method names, so I'm fine with leaving this as is.

roman-shuhov:
Thank you for feedback! Could you please fixes? 

Also have a question about generated tests. Why do we use Skip attribute for them? As I see logic is Skip = Index > 0, but not sure why.

ErikSchierboom:
Merged! Thanks. :tada: 

> Why do we use Skip attribute for them? As I see logic is Skip = Index > 0, but not sure why.

This is because Exercism aims for the user to solve the tests one by one. The first test should be the simplest case, and then each subsequent test should introduce one additional feature to slowly build up a fully working solution.

roman-shuhov:
got it, thanks!

PR #422, opened 2017-10-28
Make exercises option take single name

ErikSchierboom:
There was currently a discrepancy between the option allowed in the build.(ps1|sh) file and the generators options. This PR fixes that by only allowing a single exercise in both cases. It also removes some unused namespaces.


robkeim:
Nice simplification @ErikSchierboom, thanks!

PR #450, opened 2017-11-18
Update canonical data for All Your Base and Reverse String

robkeim:
A couple of updates from the geneator and I also manually updated the sln file to run all exercises by default instead of only allergies.


ErikSchierboom:
Looking good. Waiting for CI to become green.

ErikSchierboom:
Merged. Thanks!

PR #452, opened 2017-11-19
Use property-based testing in diamond exercise

### LINKED ISSUE #431: Use property-based testing in diamond exercise
We should convert the tests to use [FsCheck](https://github.com/fscheck/FsCheck), similar to the [C# implementation](https://github.com/exercism/csharp/blob/master/exercises/diamond/DiamondTest.cs).

ErikSchierboom:
Closes #431


robkeim:
Well done @ErikSchierboom!

PR #467, opened 2017-12-30
Add back deleted, deprecated exercises

kytrinyx:
When migrating the data from v1 to v2, we will lose all the solutions submitted to
deprecated exercises, unless they exist in the track configuration.

This adds the deprecated binary, trinary, octal, and hexadecimal exercises back,
 leaving them deprecated so that they don't actively show up as part of the track.

I've tried to follow the new project format, adding the `Project.fs` and `$slug.fsproj`, as well as a solution stub. The solution stub in particular is likely to not be valid F#, as I just looked at other exercises and made some guesses.

We can delete them from the track again after we launch v2, if we want to.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1223,6 +1223,26 @@
       ],
       "unlocked_by": "two-fer",
       "uuid": "a6082751-98ca-45dc-aeed-cdd19a8da0ca"
+    },
+    {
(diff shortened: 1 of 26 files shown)


kytrinyx:
Ah, the build is failing because I've not yet converted the test suite. I'll take a stab at that now.

ErikSchierboom:
Thanks a lot!

PR #475, opened 2018-01-15
Remove unused exercise README insert

kytrinyx:
This deletes the empty exercise README insert, and updates the exercise README template
to remove the reference to the insert.

If the track later requires some text to be repeated in all the exercise READMEs it can be
added directly to the config/exercise_readme.go.tmpl file.

See https://github.com/exercism/meta/issues/94


ErikSchierboom:
Thanks!

PR #482, opened 2018-01-22
Handle deprecated exercises

ErikSchierboom:
This PR adds support for two things:

1. Make the status filter for generators case-insensitive
1. Allow filtering on deprecated exercises (which were previously listed in the "Unimplemented" section)


robkeim:
Thanks @ErikSchierboom!

PR #521, opened 2018-03-19
config: Fix non-core exercises being marked as core

ErikSchierboom:
See #520


robkeim:
Thanks for taking care of this @ErikSchierboom!

Issue #522, opened 2018-04-01
pov: children as member causes error FS0072; I suggest annotating arg of graphToList


petertseng:
I preferred to have my Graph's internal representation of children be a Map:

```fsharp
type 'a Graph when 'a: comparison = {value: 'a; childrenByValue: Map<'a, 'a Graph>}
```

Of course, the test wants to be able to say `graph.children` and have that be an `'a Graph list`. That's perfectly reasonable; I'l add it as a member.

```fsharp
type 'a Graph when 'a: comparison = {value: 'a; childrenByValue: Map<'a, 'a Graph>} with
  member this.children: 'a Graph list = List.map snd (Map.toList this.childrenByValue)
```

ah, apparently not quite. This causes an error:

```
vagrant@ubuntu-xenial:~/fsharp/pov$ dotnet test
Build started, please wait...
/vagrant/pov/PovTest.fs(12,9): error FS0072: Lookup on object of indeterminate type based on information prior to this program point. A type annotation may be needed prior to this program point to constrain the type of the object. This may allow the lookup to be resolved. [/vagrant/pov/Pov.fsproj]
```

I see. Well, apparently a type annotation may be needed, so if line 10 of https://github.com/exercism/fsharp/blob/master/exercises/pov/PovTest.fs#L10 were to change from `let rec graphToList graph` to either of the two following choices:

```fsharp
let rec graphToList (graph: 'a Graph)
```

```fsharp
let rec graphToList (graph: Graph<'a>)
```

That appears to have everything work for me. Therefore, I simply added the annotation to my test file, and submitted my solution. So the remaining questions I had:

* Were there any disadvantages of adding these type annotations?
    * Perhaps it now forces students to make their Graph generic. Before, perhaps it was possible for them to have a `Graph` that only works on strings. I wouldn't have encouraged it, but it was possible. Adding `'a Graph` makes it impossible.
* Was there another way to achieve what I wanted? Recall that my goal was to have a different internal representation of children, and only build the `'a Graph list` on demand, because the different internal representation is more convenient.

petertseng:
EBWODP:

> so if line 12 of https://github.com/exercism/fsharp/blob/master/exercises/pov/PovTest.fs#L10

I said the wrong line but linked to the correct line. I mean line 10. Even though the error is on line 12, the place where the annotation should be added is on line 10.

petertseng:
> Were there any disadvantages of adding these type annotations?

[...]

PR #524, opened 2018-04-02
pov: Annotate type of graph in `graphToList`

petertseng:
As discussed in https://github.com/exercism/fsharp/issues/522

The advantage of this is that it allows the student to have `children`
be a member function while using a different internal representation of
children (say, a Map).

The disadvantage is that it is now **required** that `Graph` be generic;
before this annotation it would have been feasible to create a `Graph`
that only accepts String values.

We find the tradeoff acceptable because it allows increased flexibility
in `children`. The decreased flexibility (must be generic) is balanced
by the fact that it forces the student to think about how the graph
might be made generic, which seems like a good idea anyway! As a result,
potential users of such a `Graph` will have more options.

Closes https://github.com/exercism/fsharp/issues/522

### DIFF
--- a/exercises/pov/PovTest.fs
+++ b/exercises/pov/PovTest.fs
@@ -7,7 +7,7 @@ open Xunit
 
 open Pov
 
-let rec graphToList graph = 
+let rec graphToList (graph: Graph<'a>) =
     let right =
         graph.children
         |> List.sortBy (fun x -> x.value)


ErikSchierboom:
Thanks a lot @petertseng! 🎉

petertseng:
Thanks for the advice and your time!

PR #531, opened 2018-04-05
Update dependencies

ErikSchierboom:


robkeim:
Thanks @ErikSchierboom! :tada: Curious about this PR, are there new things that these updates have that you want to start using, or is this update more about making sure that we're not falling out of date?

PR #535, opened 2018-04-09
build: use .NET Core release of Cake and simplify build script

### LINKED ISSUE #528: Use .NET Core version of Cake
At the moment, we use a CoreCLR specific version of Cake in order to run Cake on .NET Core. However, with [release 0.26.0](https://cakebuild.net/blog/2018/02/cake-v0.26.0-released), Cake is now a .NET standard 2.0 application and as such should also be able to be run directly on .NET Core. We should update our build script to use Cake instead of Cake.CoreCLR, as this will allow us to remove the .NET Core 1.x dependency in our build script.

ErikSchierboom:
Closes #528.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -1,149 +1,65 @@
 <#
 .SYNOPSIS
-This is a Powershell script to bootstrap a Cake build.
+Build and test the Exercism F# exercises.
 .DESCRIPTION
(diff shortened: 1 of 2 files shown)


robkeim:
Thanks @ErikSchierboom! :tada:

PR #537, opened 2018-04-11
Bank account bug fix and improvement

jbtule:
- Original tests had actual and expected swapped.
 - Added a double check so that `closeAccount` doesn't return None.
 - Switched to base 10 decimal type since that's the industry practice.

### DIFF
--- a/exercises/bank-account/BankAccountTest.fs
+++ b/exercises/bank-account/BankAccountTest.fs
@@ -1,4 +1,4 @@
-// This file was created manually and its version is 1.0.1.
+// This file was created manually and its version is 2.0.1.
 
 module BankAccountTest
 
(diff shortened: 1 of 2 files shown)


exercises/bank-account/BankAccountTest.fs
```diff
@@ -1,4 +1,4 @@
-// This file was created manually and its version is 1.0.0.
+// This file was created manually and its version is 2.0.1.
```
ErikSchierboom: Why not just 2.0.0?

jbtule:
I’ve been using exercism to train some dev apprentices, it’s been so nice just being able to grab a problem here to demonstrate concepts!

ErikSchierboom:
Thanks a lot! Great to hear that Exercism has been useful to train apprentices.

PR #538, opened 2018-04-15
Simplify test generators

### LINKED ISSUE #529: Simplify test generator base class
Once the last generator (see #504) has been implemented, it is a good time to see which methods of the base generator class are actually used. After this analysis, we probably want to convert all virtual methods that are not overridden to plain, non-virtual methods.

Once this is done, we can then check to see if some of those converted methods can perhaps be inlined, as they won't need to be overridden anymore.

ErikSchierboom:
In this PR, I've tried to restrict the number of method a test generator can override to a minimum. This should make it easier to create new libraries, as there are less options. As a nice side-effect, it will also make writing the documentation easier :)

Closes #529.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ paket-files
 .vs/
 .vscode/
 .dotnet
+*.fsproj.user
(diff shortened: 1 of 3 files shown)


robkeim:
Cool stuff @ErikSchierboom!

PR #539, opened 2018-04-16
generators: Add docs

### LINKED ISSUE #530: Add documentation for test generators
Once the test generator API has been cleaned-up (see #529), we should add documentation to show off the various features of the test generators and how to add a new one.

ErikSchierboom:
Closes #530. Note that this documentation refers to the simplified API that is contained in PR #538.

### DIFF
--- a/docs/GENERATORS.md
+++ b/docs/GENERATORS.md
@@ -1,0 +1,314 @@
+# Test Generators
+
+Test generators allow tracks to generate tests automatically without having to write them ourselves. Each test generator reads from the exercise's `canonical data`, which defines the name of the test, its inputs, and outputs. You can read more about exercism's approach to test suites [here](https://github.com/exercism/docs/blob/master/language-tracks/exercises/anatomy/test-suites.md).
+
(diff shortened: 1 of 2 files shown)


docs/GENERATORS.md
```diff
@@ -0,0 +1,314 @@
...
+
+For many generators, the default behavior of the `GeneratorExercise` class suffices. However, for more _complex_ generators, one can override one or more of its virtual methods:
+
+### Method: RenderInput
+
+Render the input of a test method.
+
+#### Example
+
+The `Gigasecond` generator renders the input value as a (parenthesized) `DateTime` value:
```
robkeim: You didn't link to the corresponding code here (and other places below) like you did above for the Bob generator. Was that intentional?

robkeim:
Awesome description @ErikSchierboom! This should definitely help other people implementing generators in the future. I added one minor comment, but you can feel free to merge this if it was intentional.

docs/GENERATORS.md
ErikSchierboom: Yes that was intentional, as in the F# version, all the generators are in the same file. This means that to refer to a specific generator, one would have to refer to a specific line. However, each change to that generators file would require us to also change all the line-based references. This seems like a too much work for the benefits.

PR #543, opened 2018-04-29
Formatting refactor

ErikSchierboom:
Here is the big PR I've wanted to do for quite some time. Now that virtually all generators have been created, it was easier to see which code was commonly used and should thus be generalized. In this PR, I've simplified virtually all aspects of the generators. Lots of duplicate code was merged into single functions and some unused code was removed. All in all, I think the end result has our generators look quite nice.

Note: please do a merge commit to retain the individual commits (no squashing please)


generators/Common.fs
```diff
@@ -32,86 +28,10 @@ type TestFileFormat =
     | Class
 
 module Logging =
+
+    open Serilog
```
robkeim: Any reason for this `open` being inside the `module` while most other declarations are outside?

generators/Conversion.fs
```diff
@@ -0,0 +1,75 @@
+[<AutoOpen>]
```
robkeim: I didn't know about this attribute... neat!

generators/Common.fs
ErikSchierboom: You are correct that they are currently mixed. I've looked at some code, and I think the guideline is to make the open statements as local as possible: https://github.com/Microsoft/visualfsharp/blob/master/src/fsharp/FSharp.Core/array.fs#L1274

Shall I make the change?

robkeim: Ok let's standardize on that then. Not urgent to make the change but if you'd like to I'd say let's go with another PR.

robkeim:
Great work @ErikSchierboom, thanks!

ErikSchierboom:
Thanks for reviewing!

PR #558, opened 2018-06-18
maintainers: Update ErikSchierboom info

ErikSchierboom:


ErikSchierboom:
There appears to be an issue with Travis, as the exact same URL that it fails to download can be downloaded on my system.

PR #564, opened 2018-07-23
exercises: Assert precision to two decimals

### LINKED ISSUE #562: space-age: Compare values using explicit precision

ErikSchierboom:
While mentoring some space-age solutions, I found that it sometimes failed due to a missing precision specifier. This PR adds a precision check and sets the existing ones to two decimals.

Closes #562.

### DIFF
--- a/generators/Templates/_AssertEqualWithin.liquid
+++ b/generators/Templates/_AssertEqualWithin.liquid
@@ -1,1 +1,1 @@
-{{ Sut }} |> should (equalWithin 0.000000001) {{ Expected }}
+{{ Sut }} |> should (equalWithin 0.01) {{ Expected }}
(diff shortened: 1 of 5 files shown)


generators/Templates/_AssertEqualWithin.liquid
```diff
@@ -1 +1 @@
-{{ Sut }} |> should (equalWithin 0.000000001) {{ Expected }}
\ No newline at end of file
+{{ Sut }} |> should (equalWithin 0.01) {{ Expected }}
```
robkeim: What's the reason for this change? Just to not see as many 0s?

ErikSchierboom: Correct!

robkeim:
Nice, thanks @ErikSchierboom! 🎊

Fernando-Soto:
Did you bump up the version number? My old one was 1.1.0 and so was this one?

ErikSchierboom:
That is correct, as the version number is tied to the version of the canonical data.

Fernando-Soto:
So the only way to make sure that nothing in the test suite has changed from the last time I used it while mentoring in F# is to reload the test suite and student code?

ErikSchierboom:
Correct.

PR #568, opened 2018-07-30
[Leap] Make message to ignore built-in library method F# specific

robkeim:
This was originally discussed for the C# track here:
https://github.com/exercism/csharp/issues/610

Same logic applies to F# as well


robkeim:
Let's centralize the discussion for this in the csharp PR:
https://github.com/exercism/csharp/pull/612

We'll then apply the same outcome to both tracks (unless there's some really compelling reason they should have two separate ways to handle this).

PR #575, opened 2018-08-15
The nucleotideCounts function signature does not match the unit tests…

ricemery:
The nucleotideCounts function signature does not match the unit tests.  Added the Option type to wrap the Map.


ErikSchierboom:
Great catch! Thanks!

PR #580, opened 2018-09-27
Added f sharp for fun and profit in Ressource.md

MaritimeMartin:
I hope, everything went well. I messed up a little bit with the commits.


MaritimeMartin:
I think this pull request was a little bit too quick because there seems to be a second file called LEARNING.md which already has this resource named. I wonder why there are two files for the same purpose.

ErikSchierboom:
Actually, I think there is value in having the link also listed on the resources page!

ErikSchierboom:
Merged and thanks for doing this!

PR #615, opened 2018-10-04
Updating test suite for exercise: anagram 

### LINKED ISSUE #601: Update test suite for exercise: anagram
The `anagram` exercise's [test suite](https://github.com/exercism/fsharp/blob/master/exercises/anagram/AnagramTest.fs) is automatically generated using its [test generator](https://github.com/exercism/fsharp/blob/master/generators/Generators.fs). This test generator converts the `anagram`'s [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/anagram/canonical-data.json) into the test suite just mentioned.

The canonical data for the `anagram` exercise has been updated since the test suite was last generated. We should be updating the testsuite to the latest version. To do this, one should:
- Run the test generator for the exercise. This is done by opening a command prompt in the `generators` directory and running: `dotnet run -e anagram`.
- At this point, you should check to see what the updated test suite looks like. There are basically two options: one, the test generator was compatible with the updated canonical data and the test suite was updated and still has the same structure. The second option is that the test generator could not process the changed canonical data, in which case the generator has to be modified. In that case, you should check out the [generator docs](https://github.com/exercism/fsharp/blob/master/docs/GENERATORS.md) in which the generators setup is explained in detail.
- Finally, please make sure that the [example implementation](https://github.com/exercism/fsharp/blob/master/exercises/anagram/Example.fs) passes all the tests. You can verify this by opening a command prompt in the root directory and running `./build.sh --exercise=anagram` or `.\build.ps1 -Exercise anagram`, depending on your operating system.

vrnithinkumar:
Fixes #601

### DIFF
--- a/exercises/anagram/Anagram.fs
+++ b/exercises/anagram/Anagram.fs
@@ -1,3 +1,3 @@
 ﻿module Anagram
 
-let anagrams sources target = failwith "You need to implement this function."
+let findAnagrams sources target = failwith "You need to implement this function."
--- a/exercises/anagram/AnagramTest.fs
+++ b/exercises/anagram/AnagramTest.fs
@@ -1,4 +1,4 @@
-// This file was auto-generated based on version 1.2.0 of the canonical data.
+// This file was auto-generated based on version 1.3.0 of the canonical data.
 
 module AnagramTest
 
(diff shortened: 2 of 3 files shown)


robkeim:
Change looks fine to me, thanks!

PR #618, opened 2018-10-04
updating binary-search test cases

### LINKED ISSUE #595: Update test suite for exercise: binary-search
The `binary-search` exercise's [test suite](https://github.com/exercism/fsharp/blob/master/exercises/binary-search/BinarySearchTest.fs) is automatically generated using its [test generator](https://github.com/exercism/fsharp/blob/master/generators/Generators.fs). This test generator converts the `binary-search`'s [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/binary-search/canonical-data.json) into the test suite just mentioned.

The canonical data for the `binary-search` exercise has been updated since the test suite was last generated. We should be updating the test suite to the latest version. To do this, one should:
- Run the test generator for the exercise. This is done by opening a command prompt in the `generators` directory and running: `dotnet run -e binary-search`.
- At this point, you should check to see what the updated test suite looks like. There are basically two options: one, the test generator was compatible with the updated canonical data and the test suite was updated and still has the same structure. The second option is that the test generator could not process the changed canonical data, in which case the generator has to be modified. In that case, you should check out the [generator docs](https://github.com/exercism/fsharp/blob/master/docs/GENERATORS.md) in which the generators setup is explained in detail.
- Finally, please make sure that the [example implementation](https://github.com/exercism/fsharp/blob/master/exercises/binary-search/Example.fs) passes all the tests. You can verify this by opening a command prompt in the root directory and running `./build.sh --exercise=binary-search` or `.\build.ps1 -Exercise binary-search`, depending on your operating system.

vrnithinkumar:
fixes #595

### DIFF
--- a/generators/Generators.fs
+++ b/generators/Generators.fs
@@ -82,8 +82,8 @@ type BinarySearch() =
 
     let nonNegativeNumberFromNonErrorObject value =
         match Option.ofNonErrorObject value with 
-            | None -> None
-            | _ -> Option.ofNonNegativeNumber value
+        | None -> None
+        | _ -> Option.ofNonNegativeNumber value
     
     override __.PropertiesWithIdentifier _ = ["array"; "value"; "expected"]
 
--- a/generators/Generators.fs
+++ b/generators/Generators.fs
@@ -80,6 +80,11 @@ type BeerSong() =
 type BinarySearch() = 
     inherit GeneratorExercise()
 
+    let nonNegativeNumberFromNonErrorObject value =
+        match Option.ofNonErrorObject value with 
+            | None -> None
+            | _ -> Option.ofNonNegativeNumber value
+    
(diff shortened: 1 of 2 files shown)


generators/Generators.fs
```diff
@@ -80,14 +80,19 @@ type BeerSong() =
 type BinarySearch() = 
     inherit GeneratorExercise()
 
+    let nonNegativeNumberFromNonErrorObject value =
+        match Option.ofNonErrorObject value with 
+            | None -> None
```
ErikSchierboom: Tiny nit: could you have the pipes be indented at the same level as the `match` clause?

vrnithinkumar: Changed the indentation

ErikSchierboom:
Great, thanks! 🎉

Issue #621, opened 2018-10-05
generator output doesn't indicate when nothing was generated


jrr:
I typo'd an exercise name, and from the output I couldn't tell that anything had gone wrong:

```
jrr@jrrmbp ~/r/f/generators (master)> dotnet run -e collatz-conjectur
Using launch settings from /Users/jrr/repos/fsharp/generators/Properties/launchSettings.json...
[22:41:57 INF] Re-generating test classes...
[22:41:57 INF] Cloning repository...
[22:41:59 INF] Repository cloned.
[22:41:59 INF] Updating repository to latest version...
[22:41:59 INF] Updated repository to latest version.
[22:41:59 INF] Re-generated test classes.
```

It would have been helpful if there was an error, or at least a log message.

jrr:
This is fixed on master now. It now looks like this:
```
jrr@jrrmbp ~/r/f/generators (master) [130]> dotnet run -e borkborkbork
Using launch settings from /Users/jrr/repos/fsharp/generators/Properties/launchSettings.json...
[09:37:04 INF] Re-generating test classes...
[09:37:04 INF] Updating repository to latest version...
[09:37:05 INF] Updated repository to latest version.
[09:37:05 WRN] No exercises matched given options.
```

PR #622, opened 2018-10-05
generator: fail when no exercises matched the given options

jrr:
Here's an extremely simple solution to https://github.com/exercism/fsharp/issues/621

I don't know if anyone or anything was relying on the existing _generate-nothing-and-throw-no-errors_ behavior, but thought I'd start the discussion and try to improve the experience for other people that land here via Hacktoberfest. I'm totally open to other, better ways of improving this!

### DIFF
--- a/generators/Program.fs
+++ b/generators/Program.fs
@@ -49,6 +49,9 @@ let private regenerateTestClasses options =
     
     createExercises options
     |> List.filter (shouldBeIncluded options)
+    |> (function
+        | [] -> failwith "no exercises matched given options"
+        | x -> x )
     |> List.iter regenerateTestClass'
 
     Log.Information("Re-generated test classes.")
--- a/generators/Program.fs
+++ b/generators/Program.fs
@@ -51,8 +51,8 @@ let private regenerateTestClasses options =
     |> List.filter (shouldBeIncluded options)
     |> function
         | [] -> Log.Error "No exercises matched given options."
-        | x ->
-            List.iter regenerateTestClass' x
+        | exercises ->
+            List.iter regenerateTestClass' exercises
             Log.Information("Re-generated test classes.")
 
 [<EntryPoint>]
--- a/generators/Program.fs
+++ b/generators/Program.fs
@@ -50,7 +50,7 @@ let private regenerateTestClasses options =
     createExercises options
     |> List.filter (shouldBeIncluded options)
     |> function
-        | [] -> Log.Error "No exercises matched given options."
+        | [] -> Log.Warning"No exercises matched given options."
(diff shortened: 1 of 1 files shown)


generators/Program.fs
```diff
@@ -49,6 +49,9 @@ let private regenerateTestClasses options =
     
     createExercises options
     |> List.filter (shouldBeIncluded options)
+    |> (function
```
ErikSchierboom: I really like this feature! Well thought of!

I might prefer a little change to the implementation though. Perhaps you could extract the logic to a helper function, in which you can then use the `Log.Error` function to log the actual error. This is more consistent with how errors are handled elsewhere in the file.

Another option would be to do some matching:

```fsharp
let exercisesToGenerate = createExercises options |> List.filter (shouldBeIncluded options)
match exercisesToGenerate with
| [] -> Log.Error("No exercises matched given options")
| _  -> List.iter regenerateTestClass' exercisesToGenerate
```

See which ever you prefer.

generators/Program.fs
```diff
@@ -49,9 +49,11 @@ let private regenerateTestClasses options =
     
     createExercises options
     |> List.filter (shouldBeIncluded options)
-    |> List.iter regenerateTestClass'
-
-    Log.Information("Re-generated test classes.")
+    |> function
+        | [] -> Log.Error "No exercises matched given options."
+        | x ->
```
ErikSchierboom: Tiny nit: could you rename `x` to `exercises`?

generators/Program.fs
jrr: Alright, I pushed an update that's neither of those things.

I like avoiding the _"Re-generated test classes."_ message when nothing was generated, and I almost always prefer extending pipelines over declaring intermediates.

What do you think? Do you have a strong preference for one of your suggestions over this? Totally willing to change it again.

generators/Program.fs
jrr: Good idea! I could stand to develop a habit of naming those better.

generators/Program.fs
```diff
@@ -49,9 +49,11 @@ let private regenerateTestClasses options =
     
     createExercises options
     |> List.filter (shouldBeIncluded options)
-    |> List.iter regenerateTestClass'
-
-    Log.Information("Re-generated test classes.")
+    |> function
+        | [] -> Log.Error "No exercises matched given options."
```
ErikSchierboom: Maybe `Log.Warning` would be a better error level on second thought. So sorry about this.

ErikSchierboom:
Love it! Thanks a lot. 🎉

PR #623, opened 2018-10-05
Update pascals-triangle

jrr:


jrr:
https://github.com/exercism/fsharp/issues/587

ErikSchierboom:
Hi! Nice job! I do have one request though: could you split this single, multi-exercise PR into separate PR's? This makes reviewing easier and means that I can merge one even when the other is still being worked on.

As for the actual contents, the `pascals-triangle` exercise now had its error case removed. This means that the implementation should no return an `int list list option` but just a `int list list`. Could you make the necessary changes?

jrr:
Done! The option type is gone. How'd I do with the generator?

ErikSchierboom:
> Done! The option type is gone. How'd I do with the generator?

You did brilliant! I'll just wait for the CI to become green, and then I'll merge!

ErikSchierboom:
Thanks a lot! 🎉

PR #624, opened 2018-10-05
regenerate collatz-conjecture

jrr:
Here's a separate PR for just this exercise, though the change is trivial.


ErikSchierboom:
Thanks! The change is indeed trivial, but the commit history is now also more clearer than when the PR contained multiple exercises.

PR #636, opened 2018-10-09
Add generator for exercise: pythagorean-triplet

### LINKED ISSUE #584: Add generator for exercise: pythagorean-triplet
The [pythagorean-triplet](https://github.com/exercism/problem-specifications/blob/master/exercises/pythagorean-triplet/description.md) now has [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/pythagorean-triplet/canonical-data.json). We should replace the existing, manually crafted test suite with on that is automatically generated based on the canonical data.

To implement a generator for the `pythagorean-triplet` exercise, the following needs to be done:

- Create a generator to automatically convert the [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/pythagorean-triplet/canonical-data.json) to a test file. For more information on how to do this, check the [generators docs](https://github.com/exercism/fsharp/blob/master/docs/GENERATORS.md).
- Make sure the implementation file passes all the tests (hint: you can verify this by running `./build.sh --exercise=pythagorean-triplet` or `.\build.ps1 -Exercise pythagorean-triplet`.

nemesv:
Fixes #584 

The exercise have changed a little bit: now we need to search for pythagorean triplets whose sum is equal to the given parameter. So I've also changed the example implementation.

One of the test cases is very big (n = 30000) and it takes  ~1.5 min to run the test suite with the current example implementation. So any performance improvements are welcomed :)

### DIFF
--- a/exercises/pythagorean-triplet/PythagoreanTriplet.fs
+++ b/exercises/pythagorean-triplet/PythagoreanTriplet.fs
@@ -1,7 +1,3 @@
 module PythagoreanTriplet
 
-let triplet x y z = failwith "You need to implement this function."
-
-let isPythagorean triplet = failwith "You need to implement this function."
-
 let tripletsWithSum sum = failwith "You need to implement this function."
--- a/exercises/pythagorean-triplet/PythagoreanTriplet.fs
+++ b/exercises/pythagorean-triplet/PythagoreanTriplet.fs
@@ -4,4 +4,4 @@ let triplet x y z = failwith "You need to implement this function."
 
 let isPythagorean triplet = failwith "You need to implement this function."
 
-let pythagoreanTriplets min max = failwith "You need to implement this function."
+let tripletsWithSum sum = failwith "You need to implement this function."
--- a/exercises/pythagorean-triplet/PythagoreanTripletTest.fs
+++ b/exercises/pythagorean-triplet/PythagoreanTripletTest.fs
@@ -9,15 +9,15 @@ open PythagoreanTriplet
 
 [<Fact>]
 let ``Triplets whose sum is 12`` () =
-    tripletsWithSum 12 |> should equal [triplet 3 4 5]
+    tripletsWithSum 12 |> should equal [(3, 4, 5)]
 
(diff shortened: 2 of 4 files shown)


exercises/pythagorean-triplet/PythagoreanTriplet.fs
```diff
@@ -4,4 +4,4 @@ let triplet x y z = failwith "You need to implement this function."
 
 let isPythagorean triplet = failwith "You need to implement this function."
```
ErikSchierboom: This function does not need to be implemented anymore, and can thus be removed.

exercises/pythagorean-triplet/PythagoreanTripletTest.fs
```diff
@@ -1,36 +1,37 @@
...
-[<InlineData(3, 3, 3, false)>]
-[<InlineData(5, 6, 7, false)>]
-let ``Can recognize a valid pythagorean`` (x: int) (y: int) (z: int) (expected: bool) =
-    let actual = triplet x y z
-    isPythagorean actual |> should equal expected
-   
+
+[<Fact>]
+let ``Triplets whose sum is 12`` () =
+    tripletsWithSum 12 |> should equal [triplet 3 4 5]
```
ErikSchierboom: What do you think about just using a plain tuple to represent a triplet? So instead of `[triplet 3 4 5]`, it would become `[(3, 4, 5)]`? I think I have a slight preference for that version, even though the old test suite used a `triplet` function.

exercises/pythagorean-triplet/PythagoreanTriplet.fs
nemesv: OK, done.

exercises/pythagorean-triplet/PythagoreanTripletTest.fs
nemesv: I'm also fine with using plain tuples. I've updated the generator and the example implementation

ErikSchierboom:
I think these changes have improved this exercise a lot! Thanks for the hard work!

Issue #590, opened 2018-10-02
Update test suite for exercise: phone-number


ErikSchierboom:
The `phone-number` exercise's [test suite](https://github.com/exercism/fsharp/blob/master/exercises/phone-number/PhoneNumberTest.fs) is automatically generated using its [test generator](https://github.com/exercism/fsharp/blob/master/generators/Generators.fs). This test generator converts the `phone-number`'s [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/phone-number/canonical-data.json) into the test suite just mentioned.

The canonical data for the `phone-number` exercise has been updated since the test suite was last generated. We should be updating thetest suite to the latest version. To do this, one should:
- Run the test generator for the exercise. This is done by opening a command prompt in the `generators` directory and running: `dotnet run -e phone-number`.
- At this point, you should check to see what the updated test suite looks like. There are basically two options: one, the test generator was compatible with the updated canonical data and the test suite was updated and still has the same structure. The second option is that the test generator could not process the changed canonical data, in which case the generator has to be modified. In that case, you should check out the [generator docs](https://github.com/exercism/fsharp/blob/master/docs/GENERATORS.md) in which the generators setup is explained in detail.
- Finally, please make sure that the [example implementation](https://github.com/exercism/fsharp/blob/master/exercises/phone-number/Example.fs) passes all the tests. You can verify this by opening a command prompt in the root directory and running `./build.sh --exercise=phone-number` or `.\build.ps1 -Exercise phone-number`, depending on your operating system.

ErikSchierboom:
This is odd. I'll get back to you tomorrow

ErikSchierboom:
I've looked into this a bit, and it is really weird. If I run `dotnet run -e phone-number`, the test file gets updated. This works both on my two machines at home (Windows and Mac), and my work machine. Let's see if we can figure out what goes wrong.

First, could you print the console output that you get when running the command?

ErikSchierboom:
Note: the problem has been located! The instructions forgot to mention that you have to run the dotnet command in the `generators` directory. Sorry about that.

MaritimeMartin:
For the moment that are the changes in Generator.fs:

```
override __.RenderExpected (_, _, value) =
        match value.SelectToken "error" with
        | null  -> 
            value
            |> string
            |> sprintf "Ok %sUL"
        | error -> 
            error
            |> string
            |> sprintf "Error \"%s\""
        |> String.parenthesize

```

Was before:

```
[...]

PR #637, opened 2018-10-09
Updated the phone number test suit

MaritimeMartin:
This should solve #590 

The workload was quite intense for a newbie like me, but I got a solution (Yeah! :-) )

The new canonical data made it necessary to revise the whole suit, including Example.fs and the generator. In my eyes, the new data made the exercise a lot more difficult, because the error handling became immense.
I guess there is some polishing to be done.

### DIFF
--- a/exercises/phone-number/Example.fs
+++ b/exercises/phone-number/Example.fs
@@ -1,19 +1,46 @@
 ﻿module PhoneNumber
 
 open System
+open System.Numerics
+open System.Text.RegularExpressions
 
-let private validate (input: string) =
-    let isStartDigit digit = List.contains digit ['2'..'9']
+let private deleteFillers (input: string):string = 
+    input
+    |> Seq.filter (fun c -> not (List.contains c ['+';'.';'-';' ';'(';')']))
+    |> String.Concat
 
-    if isStartDigit input.[0] && isStartDigit input.[3] then
-        Some input
-    else 
-        None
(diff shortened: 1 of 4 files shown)


generators/Generators.fs
```diff
@@ -864,11 +864,22 @@ type PerfectNumbers() =
...
+        match value.SelectToken "error" with
+        | null  -> 
+            value
+            |> string
+            |> sprintf "Ok %sUL"
+        | error -> 
+            error
+            |> string
+            |> sprintf "Error \"%s\""
+        |> String.parenthesize
```
ErikSchierboom: I think you can drop this line, as the parentheses in the declaration are unneeded I believe.

exercises/phone-number/Example.fs
```diff
@@ -2,18 +2,43 @@
...
+    | i when i.[0] = '1' -> Error "area code cannot start with one"
+    | _ -> Ok input
+
+let private checkExchangeCode (input:string): Result<string, string> =
+    match input with
+    | i when i.[3] = '0' -> Error "exchange code cannot start with zero"
+    | i when i.[3] = '1' -> Error "exchange code cannot start with one"
+    | _ -> Ok input
+
+let clean (input: string): Result<uint64, string> =
```
ErikSchierboom: | love how you implemented this!

ErikSchierboom:
This is some great work! I love that we now have an exercise that properly shows using the Result type. Thanks a lot!

Issue #585, opened 2018-10-02
Add generator for exercise: sgf-parsing


ErikSchierboom:
The [sgf-parsing](https://github.com/exercism/problem-specifications/blob/master/exercises/pythagorean-triplet/description.md) now has [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/sgf-parsing/canonical-data.json). We should replace the existing, manually crafted test suite with on that is automatically generated based on the canonical data.

To implement a generator for the `sgf-parsing` exercise, the following needs to be done:

- Create a generator to automatically convert the [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/sgf-parsing/canonical-data.json) to a test file. For more information on how to do this, check the [generators docs](https://github.com/exercism/fsharp/blob/master/docs/GENERATORS.md).
- Make sure the implementation file passes all the tests (hint: you can verify this by running `./build.sh --exercise=sgf-parsing` or `.\build.ps1 -Exercise sgf-parsing`.

roman-shuhov:
I'll work on this

PR #644, opened 2018-10-20
Adding generator for sgf-parsing; small code fixes for example (#585)

roman-shuhov:


ErikSchierboom:
I'll come back to this later! Thanks for doing this.

exercises/sgf-parsing/SgfParsingTest.fs
```diff
@@ -1,98 +1,63 @@
...
-    parseSgf input |> should equal expected
+    let expected = Some (Node (Map.empty, []))
+    parseSgf "(;)" |> should equal expected   
 
 [<Fact(Skip = "Remove to run test")>]
 let ``Single node tree`` () =
-    let input = "(;A[B])"
-    let expected = Some (treeWithNoChildren [("A", ["B"])])
-    parseSgf input |> should equal expected
+    let expected = Some (Node (Map.empty.Add("A", ["B"]), []))
```
ErikSchierboom: I think I would prefer to use `Map.ofList` instead of `Map.empty.Add`.

ErikSchierboom: I think I would prefer to use `Map.ofList` instead of `Map.empty.Add`.

ErikSchierboom: I think I would prefer to use `Map.ofList` instead of `Map.empty.Add`.
```suggestion
    let expected = Some (Node (Map.ofList [("A", ["B"])], []))
```

generators/Generators.fs
```diff
@@ -1289,6 +1289,42 @@ type Series() =
...
+
+            sprintf "Node (Map.empty%s, [%s])" props children
+
+        seq {
+            if canonicalDataCase.Expected.SelectToken("properties") <> null then
+                yield sprintf "let expected = Some (%s)" (canonicalDataCase.Expected |> renderTree)
+        }
+        |> Seq.toList
+
+    override __.RenderSut (canonicalDataCase: CanonicalDataCase) =
```
ErikSchierboom: Overriding this method should not be necessary. Our implementation should follow the canonical data, should we should change the Example implementation to match the canonical data. Could you remove this override?

generators/Generators.fs
```diff
@@ -1289,6 +1289,42 @@ type Series() =
         |> Option.ofNonErrorObject
         |> Option.renderParenthesized
 
+type SgfParsing() = 
+    inherit GeneratorExercise()
+
+    override __.RenderArrange (canonicalDataCase: CanonicalDataCase) =
```
ErikSchierboom: You shouldn't have to override this method, converting the expected value should happen in the `RenderExpected` method.

generators/Generators.fs
```diff
@@ -1289,6 +1289,42 @@ type Series() =
...
+                yield sprintf "let expected = Some (%s)" (canonicalDataCase.Expected |> renderTree)
+        }
+        |> Seq.toList
+
+    override __.RenderSut (canonicalDataCase: CanonicalDataCase) =
+        let input = string <| canonicalDataCase.Input.["encoded"]
+        sprintf "parseSgf \"%s\"" input
+
+    override self.RenderExpected (canonicalDataCase, _, value) = 
+        if canonicalDataCase.Expected.SelectToken("error") <> null then
```
ErikSchierboom: There are some utility functions to dealing with errors, see e.g. the following code that is used at several places:

```fsharp
value
|> Option.ofNonErrorObject
|> Option.renderParenthesized
```

You will probably need to add an `Option.map` call in between the two functions.

roman-shuhov:
Thank you for the input! I tried to address requested changes, would you mind checking my PR again?

generators/Generators.fs
```diff
@@ -1326,6 +1326,38 @@ type Series() =
         |> Option.ofNonErrorObject
         |> Option.renderParenthesized
 
+type SgfParsing() = 
+    inherit GeneratorExercise()
+
+    override this.PropertiesWithIdentifier _ = ["expected"]
```
ErikSchierboom: One final nit (sorry about this): could you also add "input" here, to have the input also be a separate identifier?

PR #651, opened 2018-10-20
Trying to optimize build by using solution file to test (#638)

roman-shuhov:
Did couple experiment runs in travis CI. Looks like it's a 25% improvement for a build time execution: 14m30s against 11m (average numbers)


ErikSchierboom:
That's a great improvement! Thanks!

PR #658, opened 2018-10-31
fix type mismatch in exercise:grade-school starting code

x-kej:
The tests (and the original canonical data for the grade-school exercise) expect `roster` to return a list of strings, such as `["Blair"; "James"; "Paul"]`. However, the starting code in GradeSchool.fs specified the return type as a list of tuples, which means implementing the function as specified will never pass the test.

This commit just changes the function signature in the starting file to match what is expected by the tests.


ErikSchierboom:
Great catch! Thanks. 🎉

PR #671, opened 2018-12-28
add rest-api exercise

davidelettieri:
Let me know any improvements/fixes needed :)

### DIFF
--- a/config.json
+++ b/config.json
@@ -1344,10 +1344,9 @@
       "slug": "rest-api",
       "uuid": "b0aaf5ac-f94d-4995-af4a-b27068cf540e",
       "core": false,
-      "unlocked_by": null,
-      "difficulty": 0,
-      "topics": null,
-      "deprecated": false
+      "unlocked_by": "clock",
+      "difficulty": 5,
+      "topics": [ "json", "parsing" ]
     }
   ]
 }
--- a/config.json
+++ b/config.json
@@ -351,7 +351,7 @@
         "floating_point_numbers",
         "math"
       ]
-    }, 
+    },
     {
       "slug": "high-scores",
       "uuid": "c67d7118-2388-4421-baa1-cc3e4f0e2867",
@@ -1339,6 +1339,15 @@
       "difficulty": 0,
       "topics": null,
       "deprecated": true
+    },
+    {
+      "slug": "rest-api",
+      "uuid": "b0aaf5ac-f94d-4995-af4a-b27068cf540e",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 0,
+      "topics": null,
+      "deprecated": false
     }
   ]
 }
--- a/exercises/rest-api/RestApi.fs
+++ b/exercises/rest-api/RestApi.fs
@@ -2,10 +2,9 @@
 
 type RestApi(database : string) =
 
-    member this.Get (url:string) =
+    member this.Get(url: string) =
         failwith "You need to implement this function."
-    member this.Get (url:string,payload:string) =
+    member this.Get(url: string, payload: string) =
         failwith "You need to implement this function."
-
-    member this.Post (url:string,payload:string)  =
+    member this.Post(url: string, payload: string)  =
         failwith "You need to implement this function."
--- a/exercises/rest-api/RestApi.fs
+++ b/exercises/rest-api/RestApi.fs
@@ -1,0 +1,3 @@
+﻿module RestApi
+
+// TODO: implement this module
--- a/exercises/rest-api/RestApi.fs
+++ b/exercises/rest-api/RestApi.fs
@@ -1,3 +1,11 @@
 ﻿module RestApi
 
-// TODO: implement this module
+type RestApi(database : string) =
+
+    member this.Get (url:string) =
+        failwith "You need to implement this function."
+    member this.Get (url:string,payload:string) =
+        failwith "You need to implement this function."
+
+    member this.Post (url:string,payload:string)  =
+        failwith "You need to implement this function."
--- a/generators/Generators.fs
+++ b/generators/Generators.fs
@@ -1693,8 +1693,8 @@ type RestApi() =
     override __.RenderArrange canonicalDataCase =
         base.RenderArrange(canonicalDataCase) @ ["let api = RestApi(database)"]
 
-    override this.RenderSut canonicalDataCase =
+    override __.RenderSut canonicalDataCase =
         match canonicalDataCase.Property with
-            | "get" -> if canonicalDataCase.Input.ContainsKey("payload") then "api.Get (url, payload)" else "api.Get url"
-            | "post" -> "api.Post (url, payload)"
-            | _ -> base.RenderSut canonicalDataCase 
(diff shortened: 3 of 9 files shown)


ErikSchierboom:
I don't have much time at the moment to review, but I'll review in a couple of days if that's okay. Thanks for doing this!

davidelettieri:
Sure, take your time

config.json
```diff
@@ -1339,6 +1339,15 @@
       "difficulty": 0,
       "topics": null,
       "deprecated": true
+    },
+    {
+      "slug": "rest-api",
+      "uuid": "b0aaf5ac-f94d-4995-af4a-b27068cf540e",
+      "core": false,
+      "unlocked_by": null,
```
ErikSchierboom: Could you have it be unlocked by the `clock` exercise?

config.json
```diff
@@ -1339,6 +1339,15 @@
...
+      "difficulty": 0,
```
ErikSchierboom: I think a difficulty rating of `5` or `6` would be appropriate.

config.json
```diff
@@ -1339,6 +1339,15 @@
...
+      "topics": null,
+      "deprecated": false
```
ErikSchierboom: This attribute can be removed.

config.json
```diff
@@ -1339,6 +1339,15 @@
...
       "topics": null,
       "deprecated": true
+    },
+    {
+      "slug": "rest-api",
+      "uuid": "b0aaf5ac-f94d-4995-af4a-b27068cf540e",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 0,
+      "topics": null,
```
ErikSchierboom: Could you set the topics to `["json", "parsing"]`.

exercises/rest-api/RestApi.fs
```diff
@@ -0,0 +1,3 @@
+﻿module RestApi
+
+// TODO: implement this module
```
ErikSchierboom: At this difficulty level (5), we usually give users a stub implementation. However, in this case I'm not sure we should do that. @robkeim what do you think?

generators/Generators.fs
```diff
@@ -1672,4 +1672,29 @@ type Zipper() =
...
+
+    override __.RenderExpected (canonicalDataCase, key, value) =
+        match key with
+        | "expected" -> ("\"\"\"" + value.ToString(Newtonsoft.Json.Formatting.None) + "\"\"\"")
+        | _ -> base.RenderExpected (canonicalDataCase, key, value)
+
+    override __.RenderArrange canonicalDataCase =
+        base.RenderArrange(canonicalDataCase) @ ["let api = RestApi(database)"]
+
+    override this.RenderAssert canonicalDataCase =
```
ErikSchierboom: I think that perhaps you could simplify this a bit by overriding the `RenderSut` method instead of the `RenderAssert` method. Could you check to see if that is correct?

exercises/rest-api/RestApi.fs
robkeim: @ErikSchierboom what's your rationale behind not wanting to give a stub implementation for this exercise out of curiosity?

ErikSchierboom: Well, I'm not completely sure. It was more of a random blurb, so maybe we should create a stub file?

robkeim: A stub file makes the most sense to me in terms of consistency.

ErikSchierboom: Agreed. @davidelettieri could you create a stub in the implementation file?

davidelettieri: I'll do it as soon as I get back from vacation

generators/Generators.fs
```diff
@@ -1693,8 +1693,8 @@ type RestApi() =
...
         base.RenderArrange(canonicalDataCase) @ ["let api = RestApi(database)"]
 
-    override this.RenderAssert canonicalDataCase = 
+    override this.RenderSut canonicalDataCase =
         match canonicalDataCase.Property with
-        | "get" -> if canonicalDataCase.Input.ContainsKey("payload") then ["api.Get (url, payload) |> should equal expected"] else ["api.Get url |> should equal expected"]
-        | "post" -> ["api.Post (url, payload) |> should equal expected"]
-        | _ -> base.RenderAssert canonicalDataCase
\ No newline at end of file
+            | "get" -> if canonicalDataCase.Input.ContainsKey("payload") then "api.Get (url, payload)" else "api.Get url"
```
ErikSchierboom: Could you re-indent the pipes to the same level as the `match` keyword?

exercises/rest-api/RestApi.fs
```diff
@@ -1,3 +1,11 @@
 ﻿module RestApi
 
-// TODO: implement this module
\ No newline at end of file
+type RestApi(database : string) =
+
+    member this.Get (url:string) =
```
ErikSchierboom: You you remove the space after the method name and add a space before the type identifier?

exercises/rest-api/RestApi.fs
```diff
@@ -1,3 +1,11 @@
...
+        failwith "You need to implement this function."
+    member this.Get (url:string,payload:string) =
```
ErikSchierboom: You you remove the space after the method name and add a space before the type identifiers?

exercises/rest-api/RestApi.fs
```diff
@@ -1,3 +1,11 @@
...
+        failwith "You need to implement this function."
+
+    member this.Post (url:string,payload:string)  =
```
ErikSchierboom: You you remove the space after the method name and add a space before the type identifiers?

davidelettieri:
Hi Erik, thanks for reviewing again. I'll fix the formatting, however I found another issue with the new test case added a few days ago "lender owes borrower same as new loan" since the balance has no decimal digit while all other tests have it. I'm not sure if testing string against string is going to work anymore, I think it will break also the c# tests on the same exercise. 

Do you have any thoughts about that?

ErikSchierboom:
I see. I just checked the other test cases, and the "Lender has negative balance" does have an expected value of `0.0`. Having then examined the [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/rest-api/canonical-data.json#L399), I found that the new expected value is not in the same format as previous expected values: https://github.com/exercism/problem-specifications/blob/master/exercises/rest-api/canonical-data.json#L253. Would you be willing to submit a PR to the canonical data to fix this?

davidelettieri:
I changed the formatting as requested and I submitted a PR to the canonical data, let's see if it is accepted.

ErikSchierboom:
Looking great! Let's wait for the problem specifications PR to be merged, and then we can regenerate the test suite and merge this! Great work.

ErikSchierboom:
@davidelettieri The problem specifications PR was merged, so I think this is now ready to be finished by simply re-running the generator.

davidelettieri:
Done!

ErikSchierboom:
Great work! Thanks a lot

PR #672, opened 2019-01-02
Looks like tests not consistent

aivanov-oneinc:
Tests expected results contains 7 notes and contains 7 interval description for them the same time. Looks like interval has one extra item. It's impossible to have 7 intervals between 7 notes. Only 6 intervals might be between 7 notes
If I remove the last item from interval tests tun OK

### DIFF
--- a/exercises/scale-generator/ScaleGeneratorTest.fs
+++ b/exercises/scale-generator/ScaleGeneratorTest.fs
@@ -17,61 +17,61 @@ let ``Chromatic scale with flats`` () =
 
 [<Fact(Skip = "Remove to run test")>]
 let ``Simple major scale`` () =
-    interval "MMmMMMm" "C" |> should equal ["C"; "D"; "E"; "F"; "G"; "A"; "B"]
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
Hi! Great find. The test suite you mention is actually based on some shared data that you can find here: https://github.com/exercism/problem-specifications/blob/master/exercises/scale-generator/canonical-data.json

Would you mind opening an issue on that repository: https://github.com/exercism/problem-specifications/issues? That would, we can get it fixed there and all tracks can benefit from this change.

aivanov-oneinc:
@ErikSchierboom done. https://github.com/exercism/problem-specifications/issues/1425

I'll also create pull request there.

ErikSchierboom:
Thanks!

ErikSchierboom:
Seeing how https://github.com/exercism/problem-specifications/issues/1425 is closed, I'll also close this one. If you think that should not be the case, feel free to let me know @aivanov-oneinc

Issue #642, opened 2018-10-17
Check example implementation file' performance


ErikSchierboom:
It might be interesting to do some type of performance analysis of the example implementations, to see where the bottlenecks are (if there are any). For example, if one test runs for 1,5 minute, it can significantly slow down the entire build.

ErikSchierboom:
Note that for the C# track, there was a PR that improved the performance of two notoriously slow exercises: `alphametics` and `largest-series-product`: https://github.com/exercism/csharp/pull/719.

davidelettieri:
I'll work on this if that's ok, I'll already noticed in the travis logs a group of tests that take ~1 min to run, I'll investigate further and get back with more info.

ErikSchierboom:
Sure! I've done a similar PR in the C# track. You could look for inspiration there (I'm on mobile hence no link)

robkeim:
I believe this is the PR @ErikSchierboom was looking for:
https://github.com/exercism/csharp/pull/719

ErikSchierboom:
Yes that's the PR I was referring to! Thanks @robkeim!

PR #690, opened 2019-02-28
improve performance of example implementation

davidelettieri:
Regarding #642 pythagorean triplet is an exercise with some straightforward performance improvements.

### DIFF
--- a/exercises/pythagorean-triplet/Example.fs
+++ b/exercises/pythagorean-triplet/Example.fs
@@ -2,14 +2,11 @@ module PythagoreanTriplet
 
 open System
 
-let private square x = x * x
-
-let private isPythagorean (x, y, z) = square x + square y = square z
-
 let tripletsWithSum sum = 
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
@davidelettieri Looking good. Did you by any chance do some performance measurements?

davidelettieri:
Sure, on my pc before my changes it was taking 10s to run the tests, now is down to 3s

ErikSchierboom:
Great to hear! Thanks for doing this. Are there are slow tests remaining?

davidelettieri:
I think palindrome products could be improved since it's taking 9 sec on my pc to run the tests, I'll start working on it today.

However pythagorean triplets and alphametics are still slower than all the other projects.

ErikSchierboom:
Cool. I think once palindrome-products has been sped up a bit, we can consider this to be done. Alphametics and pythagorean triplets are exercises that do require some number crunching, but their performance is now good enough I feel. Awesome work!

PR #698, opened 2019-03-30
improve performance of palindrome products example implementation

### LINKED ISSUE #642: Check example implementation file' performance

davidelettieri:
This is the last exercise we agreed to improve. Let me know if any change is needed. 

Close #642

### DIFF
--- a/exercises/palindrome-products/Example.fs
+++ b/exercises/palindrome-products/Example.fs
@@ -1,23 +1,26 @@
 ﻿module PalindromeProducts
 
 let isPalindrome n = 
-    let mutable current = n
-    let mutable result = 0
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
And thanks for doing this! 🎉

davidelettieri:
> Looking good! Any performance benchmarks?

Tests run in 2 seconds on my pc against 8 of previous version.

ErikSchierboom:
Superb!

PR #699, opened 2019-04-13
build: remove shell scripts

ErikSchierboom:
This PR does a couple of things:

- It removes the `.sh` scripts. We now only have to maintain the PowerShell scripts.
- The Travis integration has been simplified.
- The build script now also has a CI build target, that does the linting.
- The boostrapping of Cake has been replaced with using their global tool.


.travis.yml
```diff
@@ -1,23 +1,20 @@
+dist: xenial
 language: csharp
-dist: trusty
```
robkeim: I don't know anything about Travis so I've got a couple of questions about this to help me learn :)

1. How did you decide on the dist and why did you change it in this PR?
2. Why is the language specified as csharp and not fsharp?

robkeim: I don't know anything about Travis so I've got a couple of questions about this to help me learn :)

1. How did you decide on the dist and why did you change it in this PR?
2. Why is the language specified as csharp and not fsharp?

ErikSchierboom: 1. The distribution had to be changed as we're now installing .NET Core SDK 2.2.203, which could not be installed on trusty.
1. I'm not sure, I think the language is just there to indicate which things come pre-installed. I'm not sure if there is an fsharp language though, so I just kept it as csharp.

ErikSchierboom:
Thanks for all the reviews @robkeim!

PR #701, opened 2019-04-13
Allow regenerating single README

### LINKED ISSUE #677: Allow generating single README
The `update-docs.sh` and `update-docs.ps1` scripts allows one to re-generate the README files. They don't, however, allow for generating a single exercise. The scripts should be modified to take an `exercise` command-line parameter (see `build.ps1` and `build.sh` for an example), and if specified, should only update a single exercise.

### LINKED ISSUE #694: Make update-docs work without having manually to checkout the problem-specifications repo
Currently, the update-docs scripts rely on the [problem-specifications repo](https://github.com/exercism/problem-specifications) to be checked out in `../problem-specifications`. This means that there is a prerequisite to getting the scripts to run, which is not ideal. What we should do is:

1. Checkout the problem-specifications repository from within the script itself (skip if already done).
1. Update the problem-specifications repository to the latest version.

ErikSchierboom:
Closes #677
Closes #694

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "problem-specifications"]
+	path = problem-specifications
+	url = https://github.com/exercism/problem-specifications.git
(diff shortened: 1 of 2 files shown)


robkeim:
Thanks @ErikSchierboom 👍

PR #717, opened 2019-07-07
Ensure PowerShell errors are propagated

ErikSchierboom:
When issue https://github.com/exercism/csharp/issues/1299 was created, I found out that there was something wrong with our build pipeline. The [CI run](https://travis-ci.org/exercism/csharp/builds/554366818?utm_source=github_status&utm_medium=notification) had an error, but Travis still thought that it was a succesful build. Apparently, one has to be explicit about returning the error code of called scripts, which is what this PR adds.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,5 +11,6 @@ paket-files
 .idea/
 .vs/
 .vscode/
+.ionide/
 .dotnet
 *.fsproj.user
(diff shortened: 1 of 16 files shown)


robkeim:
It looks like the markdown exercise now isn't working on the F# track with the latest canonical data updates.

Sorry about getting us into this mess... I updated the tests and figured I'd let the CI check to see if there were any build errors not realizing that we had this bug.

ErikSchierboom:
> Sorry about getting us into this mess... I updated the tests and figured I'd let the CI check to see if there were any build errors not realizing that we had this bug. 

No problem. I completely missed this as a reviewer, and relying on the CI makes very much sense.

ErikSchierboom:
The markdown issue is likely the same that was fixed in [this commit](https://github.com/exercism/csharp/commit/4ff720d5e6463a72a471d7115d7139fcff84caac) in the C# repository.

ErikSchierboom:
@robkeim I've fixed all the build issues.

robkeim:
Great work, thanks @EricSchierboom for getting us back into good shape! :+1:

PR #718, opened 2019-07-14
Reordered parameters to match tests.

Tom01098:
For this exercise, the tests use the ordering file, flags, pattern, but the template uses the ordering pattern, flags, file. This PR fixes it to prevent confusion.


ErikSchierboom:
Great work!

PR #726, opened 2019-08-09
 Fix generator for Palindrome Products

### LINKED ISSUE #715: Fix test generators
The following test generators are currently not working:
- [x] Allergies
- [ ] Palindrome (canonical data change [here](https://github.com/exercism/problem-specifications/commit/fbb30474df21e5de259493b010d5c502c049fa40#diff-11b95c88ab2f41a69599edd346658107))

davidelettieri:
Closes #715

### DIFF
--- a/exercises/palindrome-products/Example.fs
+++ b/exercises/palindrome-products/Example.fs
@@ -11,7 +11,7 @@ let isPalindrome n =
 
 let palindrome predicate compare startValue minFactor maxFactor = 
     if minFactor > maxFactor then
-        None
+        failwith "min must be <= max"
     else
         let allPalindromes = 
             let mutable compareValue = startValue
@@ -22,10 +22,8 @@ let palindrome predicate compare startValue minFactor maxFactor =
                         compareValue <- prod
                         yield prod, (x, y)]
 
-        match List.isEmpty allPalindromes with
-        | true -> 
-            None
-        | false ->
+        if List.isEmpty allPalindromes then (None,[])
+        else
             let value = 
                 allPalindromes 
                 |> List.map fst 
@@ -37,7 +35,7 @@ let palindrome predicate compare startValue minFactor maxFactor =
                 |> List.map snd 
                 |> List.sort
 
-            Some (value, factors)
+            (Some value, factors)
 
 let largest minFactor maxFactor = palindrome List.max (>=) 0 minFactor maxFactor
 
--- a/exercises/palindrome-products/Example.fs
+++ b/exercises/palindrome-products/Example.fs
@@ -11,7 +11,7 @@ let isPalindrome n =
 
 let palindrome predicate compare startValue minFactor maxFactor = 
     if minFactor > maxFactor then
-        failwith "min must be <= max"
+        invalidArg "minFactor" "min must be <= max"
     else
         let allPalindromes = 
             let mutable compareValue = startValue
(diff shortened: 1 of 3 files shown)


exercises/palindrome-products/Example.fs
```diff
@@ -11,7 +11,7 @@ let isPalindrome n =
 
 let palindrome predicate compare startValue minFactor maxFactor = 
     if minFactor > maxFactor then
-        None
+        failwith "min must be <= max"
```
ErikSchierboom:
```suggestion
        invalidArg "min must be <= max"
```

As we are comparing parameters here, it is probably slightly more idiomatic to throw an `ArgumentException`, for which the `invalidArg` function can be used: https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/exception-handling/the-invalidarg-function

generators/Generators.fs
```diff
@@ -871,20 +871,37 @@ type PalindromeProducts() =
...
             value.SelectToken("factors")
             |> List.mapRender toFactors
 
-        sprintf "(%d, %s)" palindromeValue factors
+        match palindromeValue with
+        | null -> sprintf "(None, %s)" factors
+        | _ -> sprintf "(Some %s, %s)" palindromeValue factors
+        
+    let isError (canonicalDataCase: CanonicalDataCase) =
+        match canonicalDataCase.Expected.Value("error") with
```
ErikSchierboom: Can this be:

```suggestion
        canonicalDataCase.Expected.Value("error") <> null
```

?

generators/Generators.fs
```diff
@@ -871,20 +871,37 @@ type PalindromeProducts() =
...
     let toPalindromeProducts (value: JToken) =
-        let palindromeValue = value.Value<int>("value")
+        let palindromeValue = value.Value("value")
         let factors = 
             value.SelectToken("factors")
             |> List.mapRender toFactors
 
-        sprintf "(%d, %s)" palindromeValue factors
+        match palindromeValue with
+        | null -> sprintf "(None, %s)" factors
```
ErikSchierboom: I think you should be able to use the `Obj.render` function here.

generators/Generators.fs
```diff
@@ -871,20 +871,37 @@ type PalindromeProducts() =
...
+        | _ -> sprintf "(Some %s, %s)" palindromeValue factors
```
ErikSchierboom: I think you should be able to use the `Obj.render` function here.

davidelettieri:
Thanks for your review Erik! I'll take a look at it in a week or so.

ErikSchierboom:
Sure! Take your time.

exercises/palindrome-products/Example.fs
davidelettieri: Sure, invalidArg wants the argument as first parameter it could be the tuple (min, max)

generators/Generators.fs
davidelettieri: I think we can also go with "(None, [ ])" since it will be always like this in this case

generators/Generators.fs
davidelettieri: I'll look into it, just applying the function to the tuple (palindromeValue, factors) does not produce a good result.

generators/Generators.fs
davidelettieri: Sure, thanks!

generators/Generators.fs
ErikSchierboom: Yeah, that would work!

generators/Generators.fs
ErikSchierboom: Ah, if you don't get it working, then feel free to ignore!

davidelettieri: I didn't find a clean and easy way to use Obj.render

ErikSchierboom:
LGTM! Thanks @davidelettieri

PR #728, opened 2019-08-21
Adjust core track to help ease bottlenecks

kytrinyx:
I wrote a script that takes some data points into account to move certain core exercises to be optional side exercises, and which reorders the remaining exercises.

This is not meant to be the "be all end all" of perfection :-)
This is an attempt to make some simple changes in the short term. In the longer term, over the next 12 months, the product team will be doing a bunch of work to dig into what makes great Exercism exercises, and how to structure tracks to provide a better experience for both learners and mentors.

**Once this goes live** we will continue to monitor the continuation rate, the median wait time, and a few other stats as well, to help decide whether or not we need to take any other immediate steps. That said, if you see anything weird or worrying once the change is out, please ping us on Slack (or here) so we can discuss a fix posthaste!

**The most helpful thing you can do in reviewing this** is to look at the results and tell me whether anything is obviously terrible or suspiciously weird. I know very little about the F# track (or the language, for that matter), so I could easily have missed something.

I'd also appreciate it if you posted this to the mentors as a heads up, and also because mentors tend to have a great gut sense about the core exercises.

### Removal from core:

TL;DR: only `sum-of-multiples` was moved out of core.

This is a list of exercises that typically are unappealing to students and not very interesting to mentor. They tend to be math problems or implementations of CS algorithms and the like.

In this track the only core exercises that were slated for removal were `sum-of-multiples` and `accumulate`. I discussed with @ErikSchierboom and he mentioned that `accumulate` is useful for learning about tail recursion. In the longer term we should probably replace accumulate with a more compelling exercise aimed at tail recursion, but in the meanwhile `accumulate` should stay in core.

### Reordering:

The data that I based the reordering on was taken from the production database, and is based on the last 6 months.

I reordered by taking the original order, and then giving "penalty points" for two different things

- low continuation rates (percentage)
- long wait times in the mentor queue (median wait time in minutes)

"Continuation" is when someone completes an exercise, but then does not submit the following exercise. We have other interesting numbers that we may use to adjust things in the future, but this one seemed like our best bet for now for the bottlenecks. I somewhat arbitrarily chose 70% as the cutoff for whether or not to give penalty points, i.e. continuation rate < 70% gets penalized. The only exercise I did not penalize based on this rate was the exercise directly following `hello-world`, since `hello-world` is basically just a taste test, and it is natural to assume that many people will decide that Exercism is not really for them.

<details>
 <summary>Continuation rates</summary>

_Note that exercises with continuation rates above 100 were previously moved from side exercises to core track._

```
- hello-world (100%)
- leap (77%)
- bob (49%)
- sum-of-multiples (66%)
- two-fer (81%)
- space-age (81%)
- raindrops (87%)
- accumulate (83%)
- grade-school (90%)
- beer-song (88%)
- queen-attack (76%)
- kindergarten-garden (90%)
- clock (92%)
- robot-simulator (88%)
- allergies (105%)
- ocr-numbers (73%)
- tree-building (88%)
- pig-latin (64%)
```

</details>

Long wait times in the mentor queue is an indication that mentors don't enjoy mentoring the exercise, or put it off. This could be for a number of reasons, which we have not explicitly identified, but we've observed that sometimes the exercise is simply too difficult in the current position, which means that mentors have to go back and forth a whole bunch with students to get something right, and the discussions can be quite frustrating. Also sometimes the exercise is just not very interesting to mentor. Or if the exercise is too early in the track sometimes people will submit lots of really complicated solutions rather than a small handful of mostly reasonable solutions, also making it harder to mentor.

So this pushes exercises with long wait times further back in the core track. We may decide that we need to remove some exercises from core altogether, if wait times continue to be bad.

<details>
 <summary>Wait times</summary>

```
- hello-world (0 min)
- leap (3560 min)
- bob (2511 min)
- sum-of-multiples (307 min)
- two-fer (463 min)
- space-age (2387 min)
- raindrops (1278 min)
- accumulate (235 min)
- grade-school (2583 min)
- beer-song (3753 min)
- queen-attack (138 min)
- kindergarten-garden (3848 min)
- clock (841 min)
- robot-simulator (3022 min)
- allergies (10227 min)
- ocr-numbers (13997 min)
- tree-building (112681 min)
- pig-latin (15732 min)
```
</details>

#### Before
```
- hello-world
- leap
- bob
- sum-of-multiples
- two-fer
- space-age
- raindrops
- accumulate
- grade-school
- beer-song
- queen-attack
- kindergarten-garden
- clock
- robot-simulator
- allergies
- ocr-numbers
- tree-building
- pig-latin
```

#### After
```
- hello-world
- two-fer (-3)
- leap (+1)
- accumulate (-4)
- raindrops (-2)
- space-age
- queen-attack (-4)
- grade-school (-1)
- clock (-4)
- bob (+7)
- beer-song (+1)
- kindergarten-garden
- robot-simulator (-1)
- allergies (-1)
- ocr-numbers (-1)
- pig-latin (-2)
- tree-building
```

### DIFF
--- a/config.json
+++ b/config.json
@@ -44,18 +44,6 @@
       ]
     },
     {
-      "slug": "sum-of-multiples",
-      "uuid": "e702b75e-4c9e-40ef-bcb1-674a87222c23",
-      "core": false,
-      "unlocked_by": null,
-      "difficulty": 1,
-      "topics": [
-        "lists",
-        "math",
-        "transforming"
-      ]
-    },
-    {
       "slug": "two-fer",
       "uuid": "2ee3cc7a-db3f-4668-9983-ed6d0fea95d1",
       "core": true,
@@ -108,31 +96,6 @@
       "topics": [
         "maps",
         "sorting"
-      ]
-    },
-    {
-      "slug": "collatz-conjecture",
-      "uuid": "8b66e691-508f-40d9-8cee-1d6aec27366e",
-      "core": false,
-      "unlocked_by": "space-age",
-      "difficulty": 2,
-      "topics": [
-        "algorithms",
-        "control_flow_conditionals",
-        "control_flow_loops",
-        "integers",
-        "math"
-      ]
-    },
-    {
-      "slug": "armstrong-numbers",
-      "uuid": "e7f74f6c-16ea-40d8-94f7-7034bc6ee938",
-      "core": false,
-      "unlocked_by": "space-age",
-      "difficulty": 2,
-      "topics": [
-        "integers",
-        "math"
       ]
     },
     {
@@ -224,6 +187,54 @@
       ]
     },
     {
+      "slug": "pig-latin",
+      "uuid": "a002e3db-8e9c-4cbf-b00f-f2090bae5d5a",
+      "core": true,
+      "unlocked_by": null,
+      "difficulty": 6,
+      "topics": [
+        "strings",
+        "transforming"
+      ]
+    },
+    {
+      "slug": "sum-of-multiples",
+      "uuid": "e702b75e-4c9e-40ef-bcb1-674a87222c23",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
+        "lists",
+        "math",
+        "transforming"
+      ]
+    },
+    {
+      "slug": "collatz-conjecture",
+      "uuid": "8b66e691-508f-40d9-8cee-1d6aec27366e",
+      "core": false,
+      "unlocked_by": "space-age",
+      "difficulty": 2,
+      "topics": [
+        "algorithms",
+        "control_flow_conditionals",
+        "control_flow_loops",
+        "integers",
+        "math"
+      ]
+    },
+    {
+      "slug": "armstrong-numbers",
+      "uuid": "e7f74f6c-16ea-40d8-94f7-7034bc6ee938",
+      "core": false,
+      "unlocked_by": "space-age",
+      "difficulty": 2,
+      "topics": [
+        "integers",
+        "math"
+      ]
+    },
+    {
       "slug": "binary-search-tree",
       "uuid": "4e786e56-2658-445f-ac91-64dd9c38dbb3",
       "core": false,
@@ -233,17 +244,6 @@
         "optional_values",
         "searching",
         "trees"
-      ]
-    },
-    {
-      "slug": "pig-latin",
-      "uuid": "a002e3db-8e9c-4cbf-b00f-f2090bae5d5a",
-      "core": true,
-      "unlocked_by": null,
-      "difficulty": 6,
-      "topics": [
-        "strings",
-        "transforming"
       ]
     },
     {
--- a/config.json
+++ b/config.json
@@ -46,7 +46,7 @@
     {
       "slug": "sum-of-multiples",
       "uuid": "e702b75e-4c9e-40ef-bcb1-674a87222c23",
-      "core": true,
+      "core": false,
       "unlocked_by": null,
       "difficulty": 1,
       "topics": [
@@ -114,7 +114,7 @@
       "slug": "collatz-conjecture",
       "uuid": "8b66e691-508f-40d9-8cee-1d6aec27366e",
       "core": false,
-      "unlocked_by": "sum-of-multiples",
+      "unlocked_by": "space-age",
       "difficulty": 2,
       "topics": [
         "algorithms",
@@ -128,7 +128,7 @@
       "slug": "armstrong-numbers",
       "uuid": "e7f74f6c-16ea-40d8-94f7-7034bc6ee938",
       "core": false,
-      "unlocked_by": "sum-of-multiples",
+      "unlocked_by": "space-age",
       "difficulty": 2,
       "topics": [
         "integers",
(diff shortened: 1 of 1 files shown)


kytrinyx:
Ah! I forgot to take into account that if we remove exercises from core, we must have a different exercise unlock the associated side exercises. I will fix my script and re-run it.

kytrinyx:
FYI I made a few adjustments to the script, and regenerated the PR. The PR description has been updated to reflect the final order.

config.json
```diff
@@ -22,48 +22,47 @@
       ]
     },
     {
-      "slug": "leap",
-      "uuid": "66d974b5-18fc-4993-b5f2-7beda4f4afa3",
+      "slug": "two-fer",
```
ErikSchierboom: This change was long overdue, `two-fer` should really be the second exercise.

config.json
```diff
@@ -22,48 +22,47 @@
...
-        "control_flow_if_else_statements",
-        "integers"
+        "optional_values",
+        "strings"
       ]
     },
     {
-      "slug": "bob",
-      "uuid": "b3c4d578-10c8-47bc-b0ae-149ed8da530a",
+      "slug": "leap",
```
ErikSchierboom: It will be really interesting to see what happens to leap's continuation rate (hopefully it increases). Let's keep monitoring this.

config.json
```diff
@@ -22,48 +22,47 @@
...
       "topics": [
         "control_flow_if_else_statements",
-        "strings"
+        "integers"
       ]
     },
     {
-      "slug": "sum-of-multiples",
-      "uuid": "e702b75e-4c9e-40ef-bcb1-674a87222c23",
+      "slug": "accumulate",
```
ErikSchierboom: I both like and fear having this exercise early. The thing I like is that it introduces a concept (recursion) that is vital to the track. The implementation itself is also quite simple, but there is a nasty minor detail (tail recursion) hiding in there. We could perhaps instruct mentors to not mention tail recursion, although this is the only exercise we have that teaches this I think.

The thing I'm concerned about, is that the concept and it's application might be difficult for people to grasp. Also, the stub definition introduces generic types (`let accumulate (func: 'a -> 'b) (input: 'a list): 'b list = ...`), so that could be a bit daunting to read. This could all very well not be true, and I wouldn't want to put it after `space-age`, as that exercise is slightly more involved to solve. So the remaining option is to swap it with `raindrops`, of which I'm not sure that improves the situation a lot. For now, let's keep it here and check regularly to see if this becomes a new blocking point. Maybe people have less problems with this than I anticipate.

config.json
```diff
@@ -78,25 +77,13 @@
...
-      "difficulty": 2,
-      "topics": [
-        "filtering",
-        "text_formatting"
-      ]
-    },
-    {
-      "slug": "accumulate",
-      "uuid": "e7085050-1611-4773-9032-0e0ffb56c20e",
+      "slug": "queen-attack",
```
ErikSchierboom: This is a great catch by the algorithm! `queen-attack` introduces a very core concept (tuples) and is quite simple to implement, so it moving up by quite a bit is great. Considering its relatively simple solution, we could even move this up even more, perhaps even before `accumulate`. But once again, let's not get carried away and just see what the data will say in a couple of weeks/months.

config.json
```diff
@@ -111,28 +98,25 @@
...
+        "time"
       ]
     },
     {
-      "slug": "armstrong-numbers",
-      "uuid": "e7f74f6c-16ea-40d8-94f7-7034bc6ee938",
-      "core": false,
-      "unlocked_by": "sum-of-multiples",
-      "difficulty": 2,
+      "slug": "bob",
```
ErikSchierboom: Great to see it remove much further back.

config.json
```diff
@@ -212,6 +175,17 @@
...
+      "core": true,
+      "unlocked_by": null,
+      "difficulty": 6,
+      "topics": [
+        "strings",
+        "transforming"
+      ]
+    },
     {
       "slug": "tree-building",
```
ErikSchierboom: This is a refactoring exercise, so personally I'd be happy to see it removed. I'm not sure if we had consensus on removing refactoring exercises though. Definitely the right move to put this at the very end.

config.json
kytrinyx: Let's keep a very sharp eye on this one. I think it only moved forward this much because of the extremely low wait time on it. If you think it's worth swapping it with raindrops now, let's do that manually in this PR!

config.json
kytrinyx: Interesting. I think I would trust your gut instinct more than my script right now. I like that the script gives us indicators, and a concrete thing to discuss.

My suggestion, I think, is to move raindrops and queen attack before accumulate.

config.json
kytrinyx: Yeah, I like refactoring exercises but they're a real pain in the butt to mentor, because refactoring is such a skill in and of itself. We could argue that by the end of the track you should have a decent feel for what idiomatic code should look like in the language... but also, I'm not sure how much I believe that yet :)

Let's keep an eye on it for a few weeks and move it out of core if we still think it's a pain where it is now.

config.json
ErikSchierboom: Let's do that then!

robkeim:
Super cool that you're using data to help determine this order!

ErikSchierboom:
Thanks for the super quick review!

Issue #736, opened 2019-09-29
Benchmarking for the Tree Building exercise


TeaDrivenDev:
I'm working on improving my ability to mentor the Tree Building exercise, and the instructions mention the "slow" original code, which seems to imply that improved performance is one of the goals (they don't say any more about that, though). I have set up benchmarking for myself using BenchmarkDotNet to be able to compare a given solution with the original code, as well as a very fast implementation. I wonder if it makes sense to try to bring that to the users "in the box" with the exercise (minus the fast solution), so they can see for themselves how they're doing, as they can with the unit tests.

It would mean adding another dependency and making the project more complex than the others; also, an easy way for the user to run the benchmarks would have to be determined.

ErikSchierboom:
Great issue! There are several things here:

1. I'm not sure the original code is actually slow. Maybe we should be making it slow, especially if it is in the instructions. Did you check to see if the original code is indeed slow? If it isn't very slow, maybe we should make it slower :)
1. We don't teach users about libraries, but the performance suggestion should be approached by using `BenchmarkDotNet`, I agree.
1. The `tree-building` exercise is actually quite hard for both students and mentors. It might make sense to remove it as a core exercise and make it a side exercise. If we do this, I then also don't mind adding the additional `BenchmarkDotNet` dependency.

What do you think?

TeaDrivenDev:
Well, it's not extremely slow, but slow enough that performance improvements are possible without huge amounts of in-depth F# knowledge; many solutions seem to be at least somewhat better, just from writing more sensible code (some are also a lot slower, though). The fastest one I've seen is currently my own, after taking some hints from the previously fastest and then specifically tuning for speed; I get a 64% performance improvement over the original code, using all the unit test cases as the benchmark set. 

There may be ways to make the original code even slower, but what I've tried without making it too contrived or rewriting it only added about 10% to the running time.

I also agree that it's a difficult exercise for both sides, mostly due to its open-endedness. Removing/replacing it as a core exercise sounds like a good idea.

TeaDrivenDev:
Closing this, as it has been implemented.

PR #739, opened 2019-10-04
Add benchmarking for Tree Building exercise

TeaDrivenDev:
This is what user-runnable benchmarking for the Tree Building exercise could look like, as discussed in #736. The original code (in `TreeBuildingBaseline.fs`) and the student's work (in `TreeBuilding.fs`) are benchmarked using the input data for all the unit tests.

A few things here:
- To run the benchmarks for the two solutions with the same logic, the `Record` type had to be extracted into its own module `TreeBuildingTypes` that is accessible from all other modules. This does not add any new restrictions, as the type already had to remain unchanged before, because it was used in the unit tests.
- The test file should probably get a new version; I'm not sure what it should be, though. 1.1.0?
- The instructions should probably be expanded, elaborating more on the performance "requirement" and potentially the benchmark.
- I think BenchmarkDotNet allows changing the columns displayed in the table. It might make sense to hide most of the default columns, as they're not helpful for our use case.


ErikSchierboom:
I really like this! It's great to have at least one exercise that focuses on performance, and the setup you've made works well. 

> To run the benchmarks for the two solutions with the same logic, the Record type had to be extracted into its own module TreeBuildingTypes that is accessible from all other modules.

This is fine, and in fact teaches the student about sharing types between files :)

> The test file should probably get a new version; I'm not sure what it should be, though. 1.1.0?

I'm thinking this being a very different version of the exercise we should be bumping the version to 2.0.0 perhaps?

> The instructions should probably be expanded, elaborating more on the performance "requirement" and potentially the benchmark.

Updating the README is a great idea. I'm thinking perhaps adding some comments in the code too, for the students that don't read the README (that's me ;))?

> I think BenchmarkDotNet allows changing the columns displayed in the table. It might make sense to hide most of the default columns, as they're not helpful for our use case.

Feel free to look into this!

Overall I love this PR! Thanks a ton for doing this.

Note to self: before merging this, we'll need to remove this exercise from the core exercises.

TeaDrivenDev:
- Reduced summary columns to `Mean` and `Allocated`
- Added/updated file version information
- Expanded instructions
- Slight performance "deprovement"

exercises/tree-building/README.md
```diff
@@ -1,5 +1,21 @@
 # Tree Building
 
+This exercise is different from most others in that you already start with
```
ErikSchierboom: Great introduction!

exercises/tree-building/TreeBuilding.fs
```diff
@@ -1,26 +1,35 @@
-﻿module TreeBuilding
+﻿// This is the file you need to modify for your own solution.
+// The unit tests will use this code, and it will be used by the benchmark tests
+// for the "Mine" row of the summary table.
 
-type Record = { RecordId: int; ParentId: int }
-type Tree = 
+// Remember to not only run the unit tests for this exercise, but also the
+// benchmark tests using `dotnet run -c Release`.
+// Please refer to the instructions for more information about the benchmark tests.
```
ErikSchierboom: Excellent addition.

ErikSchierboom:
With #742 merged, this has now also been merged. Thanks a ton @TeaDrivenDev! Really excited about this change. 🎉

robkeim:
Cool stuff @TeaDrivenDev!

PR #742, opened 2019-10-08
Remove Tree Building from core exercises

TeaDrivenDev:
This removes "Tree Building" from the core exercises on this track, to be able to go ahead with #739.

I'm wondering if it should be replaced by a similarly difficult exercise from the "side" pool, as this leaves the track with only 16 core exercises, which I feel isn't a lot, but I don't know the other exercises well enough to suggest one.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1049,7 +1049,7 @@
       "slug": "sublist",
       "uuid": "d17b0c35-49a7-4ea6-b68c-7b2b56dc5968",
       "core": false,
-      "unlocked_by": "tree-building",
+      "unlocked_by": "pig-latin",
       "difficulty": 7,
       "topics": [
         "lists",
--- a/config.json
+++ b/config.json
@@ -1129,7 +1129,7 @@
       "slug": "hangman",
       "uuid": "ab0f1b66-011f-4aa3-89c5-c89427121279",
       "core": false,
-      "unlocked_by": "tree-building",
+      "unlocked_by": "pig-latin",
       "difficulty": 8,
       "topics": [
         "reactive_programming"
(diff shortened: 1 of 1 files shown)


TeaDrivenDev:
Okay, this fails because a few other exercises are set up to be unlocked by this one. I saw that when making the change, but thought it might not matter.

Do we just remap those to being unlocked by "Pig Latin" instead?

ErikSchierboom:
> Do we just remap those to being unlocked by "Pig Latin" instead?

Yep!

ErikSchierboom:
@TeaDrivenDev The build still fails unfortunately. Could you take a look?

TeaDrivenDev:
@ErikSchierboom I'm not sure I can do much here. I ran the `test.ps1` script on my machine, and it runs for a while and outputs a lot of things, and then eventually stops with

```
Test Run Successful.
Total tests: 19
     Passed: 19
 Total time: 2,6347 Seconds
```

without dropping back to a prompt. As I don't fully understand what's happening here and have no working knowledge of PowerShell, I don't know how I would go about debugging this, or if that even makes sense.

ErikSchierboom:
Weird. I'll take a look.

ErikSchierboom:
@TeaDrivenDev I've found the issue. There is still one exercise (`sublist`) that is being unlocked by `tree-building`:

```
The exercise 'sublist' is being unlocked by a non-core exercise. Non-core exercises can only be unlocked by core exercises.
```

TeaDrivenDev:
@ErikSchierboom Okay, fixed. Odd though that it didn't come up properly in the build errors as the other ones did.

ErikSchierboom:
Great work! Thanks.

PR #749, opened 2019-11-03
Add test that will fail for non-tail-recursive implementations

connorads:
It's a bit hard to know if an implementation is tail-recursive [without looking at the MSIL](https://devblogs.microsoft.com/fsharpteam/tail-calls-in-f/).

This test should cause a `StackOverflowException` to be thrown when `accumulate` has a non-tail recursive implementation.
image
Which unfortunately takes down the test host but I guess that's not the end of the world.

When `accumulate` has a tail-recursive implementation then the test passes
image

And the test run time is similar without this new test (thinking that the new test might be slow)
image

### DIFF
--- a/exercises/accumulate/AccumulateTest.fs
+++ b/exercises/accumulate/AccumulateTest.fs
@@ -35,3 +35,8 @@ let ``Accumulate reversed strings`` () =
 let ``Accumulate within accumulate`` () =
     accumulate (fun (x:string) -> String.concat " " (accumulate (fun y -> x + y) ["1"; "2"; "3"])) ["a"; "b"; "c"]
     |> should equal ["a1 a2 a3"; "b1 b2 b3"; "c1 c2 c3"]
+
+[<Fact(Skip = "Remove to run test")>]
+let ``Accumulate large data set without stack overflow`` () =
+    accumulate id [1..100000]
+    |> should equal [1..100000]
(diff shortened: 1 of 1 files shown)


exercises/accumulate/AccumulateTest.fs
```diff
@@ -34,4 +34,9 @@ let ``Accumulate reversed strings`` () =
 [<Fact(Skip = "Remove to run test")>]
 let ``Accumulate within accumulate`` () =
     accumulate (fun (x:string) -> String.concat " " (accumulate (fun y -> x + y) ["1"; "2"; "3"])) ["a"; "b"; "c"]
-    |> should equal ["a1 a2 a3"; "b1 b2 b3"; "c1 c2 c3"]
\ No newline at end of file
+    |> should equal ["a1 a2 a3"; "b1 b2 b3"; "c1 c2 c3"]
+
+[<Fact(Skip = "Remove to run test")>]
+let ``Accumulate large data set without stack overflow`` () =
```
ErikSchierboom: I like it! The only thing I'd love to see is for the version at the top of the file to be incremented. Could you update that version?

connorads: No problemo. I incremented the major version from `1.0.0` to `2.0.0`
(_because that's what other people seemed to have done in this repo_ 😸 )

ErikSchierboom:
Lovely work. Thanks a lot! Merged 🎉

PR #767, opened 2020-01-12
Update dnd character exercise

### LINKED ISSUE #764: "D&D Character" requires using a class
The solution stub for the "D&D Character" exercise contains a class for the type to be implemented, and the unit tests directly call the constructor, which leaves the student no option than to actually solve this using a class. This is not preferred in F#, though, and there's no good reason why using a class would make sense here in any way.

The tests should just call a creation function here and not prescribe an implementation. I'm not sure what to do about the solution stub, though.

davidelettieri:
Close #764, I hope this is a satisfying implementation, if not I'll be happy to change whatever needed :)

I also updated another small detail of the generator since the test "Random ability is within range" was creating the value of ability once and testing in ten times, intead of creating it 10 times and testing once per value.

### DIFF
--- a/exercises/dnd-character/DndCharacter.fs
+++ b/exercises/dnd-character/DndCharacter.fs
@@ -8,5 +8,7 @@ let modifier x =
 let ability() = 
     failwith "You need to implement this function."
 
+// TODO: define the Character type
+
 let mkCharacter() =
     failwith "You need to implement this function."
--- a/exercises/dnd-character/DndCharacter.fs
+++ b/exercises/dnd-character/DndCharacter.fs
@@ -10,5 +10,5 @@ let ability() =
 
 // TODO: define the Character type
 
-let mkCharacter() =
+let createCharacter() =
     failwith "You need to implement this function."
--- a/exercises/dnd-character/DndCharacter.fs
+++ b/exercises/dnd-character/DndCharacter.fs
@@ -8,12 +8,5 @@ let modifier x =
 let ability() = 
     failwith "You need to implement this function."
 
-type DndCharacter() =
-    member __.Strength with get() = failwith "You need to implement this function."
-    member __.Dexterity with get() = failwith "You need to implement this function."
-    member __.Constitution with get() = failwith "You need to implement this function."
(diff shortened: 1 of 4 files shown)


exercises/dnd-character/DndCharacterTest.fs
```diff
@@ -73,15 +73,15 @@ let ``Ability modifier for score 18 is +4`` () =
 
 [<Fact(Skip = "Remove to run test")>]
 let ``Random ability is within range`` () =
-    let ability = ability()
     for i in 1 .. 10 do
+        let ability = ability()
```
robkeim: Nice catch here :+1:

ErikSchierboom: Indeed! Well caught!

exercises/dnd-character/DndCharacter.fs
```diff
@@ -8,12 +8,7 @@ let modifier x =
...
-    member __.Strength with get() = failwith "You need to implement this function."
-    member __.Dexterity with get() = failwith "You need to implement this function."
-    member __.Constitution with get() = failwith "You need to implement this function."
-    member __.Intelligence with get() = failwith "You need to implement this function."
-    member __.Wisdom with get() = failwith "You need to implement this function."
-    member __.Charisma with get() = failwith "You need to implement this function."
-    member __.Hitpoints with get() = failwith "You need to implement this function."
+// TODO: define the Character type
 
+let mkCharacter() =
```
TeaDrivenDev: Maybe `createCharacter ()`? There's no reason to abbreviate things here.

TeaDrivenDev:
Looks good, except for the function name that I'd change.

exercises/dnd-character/DndCharacter.fs
davidelettieri: You're right, I'll update accordingly

davidelettieri:
I renamed the function as requested

robkeim:
Merging this given @TeaDrivenDev's comment has been addressed.

Thanks @davidelettieri for the contribution! :tada:

PR #778, opened 2020-02-26
Rename test classes to ClassNameTests (plural) instead of ClassNameTest (singular)

robkeim:


robkeim:
I looked for that file, but didn't find it the first time... I must have clicked on the wrong directory or something. Updated now!

ErikSchierboom:
Awesome work! Thanks @robkeim!

PR #798, opened 2020-05-21
Add a test that 2500 robots have different names...

MagnusMikkelsen:
2500 names gives 99% chance of collisions if no measures have been taken to prevent it, according to the birthday problem.

[Birthday paradox explained, with calculators](https://betterexplained.com/articles/understanding-the-birthday-paradox/)

(Bear with me, this is my first PR)

### DIFF
--- a/config.json
+++ b/config.json
@@ -381,8 +381,8 @@
       "slug": "robot-name",
       "uuid": "3fbd466a-caf4-48ac-9a1e-796bc406ff1e",
       "core": false,
-      "unlocked_by": "hello-world",
-      "difficulty": 3,
+      "unlocked_by": "raindrops",
+      "difficulty": 4,
       "topics": [
         "randomness",
         "strings"
--- a/exercises/robot-name/Example.fs
+++ b/exercises/robot-name/Example.fs
@@ -8,15 +8,24 @@ let letters = ['A'..'Z']
 let digits = ['0'..'9']
 
 let NumberOfLetters = 2;
-let NumberOfDigits = 3    
+let NumberOfDigits = 3  
+
+let mutable usedNames = Set.empty<string>
 
 let takeRandomElements xs length = List.init length (fun _ -> List.item (random.Next(List.length xs)) xs)
 let generateRandomString chars length = new System.String(takeRandomElements chars length |> List.toArray)
 let generateLetters() = generateRandomString letters NumberOfLetters
 let generateDigits() = generateRandomString digits NumberOfDigits
 let generateName() = generateLetters() + generateDigits()
+let generateUniqueName() = 
+    let nextName = 
+        Seq.initInfinite (fun _ -> generateName())
+        |> Seq.filter (fun name -> not (usedNames.Contains name))
+        |> Seq.item 0
+    usedNames <- usedNames.Add(nextName)
+    nextName
 
-let mkRobot() = { name = generateName() }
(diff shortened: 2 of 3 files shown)


MagnusMikkelsen:
Hmm, the instructions for the "Robot Name" says:

> Using random names means a risk of collisions. Your solution must ensure that every existing robot has a unique name.

So I thought I'd add a test that ensures this, by generating 2500 robots and check that they have different names - only to find out that the solution checked in in this repo - doesn't fulfill this requirement 🤔

jrr:
Thanks for your contribution!

Good catch that the exercise's description didn't match its implementation re: collisions.

This could be resolved by either changing the description or changing the implementation. I favor changing the implementation (like you did) to make the exercise more interesting and educational.

Another thing to consider is what this means for the exercise's metadata, as specified in [config.json](https://github.com/exercism/fsharp/blob/master/config.json):

https://github.com/exercism/fsharp/blob/6594385d4132c08a1a7e7621966e9b20ca70c34b/config.json#L380-L390

Do we want to specifically point the student toward sequences? If so, then we could:
- Add `sequences` to the topic list
- Add some language and a link in the readme pointing the student toward sequences
- Adjust the difficulty and/or position in the track. (Is it still a "3"? Does it still make sense to unlock it after hello world?)

Are there other approaches to the problem we might point them toward instead of sequences?

Any other maintainers want to chime in?

MagnusMikkelsen:
> Are there other approaches to the problem we might point them toward instead of sequences?

Both in the solution example, and in the new test - the implementation could probably be done with loops instead. Or maybe recursion?

But afaik, keeping track of names, both in the example and in the test, requires some sort of collection.

Should I try to implement it using loops instead?

ErikSchierboom:
> This could be resolved by either changing the description or changing the implementation. I favor changing the implementation (like you did) to make the exercise more interesting and educational.

We should definitely change the implementation, as the description will be automatically generated based on data in the problem-specifications repo.

> Another thing to consider is what this means for the exercise's metadata
>
> Adjust the difficulty and/or position in the track. (Is it still a "3"? Does it still make sense to unlock it after hello world?)

I think this exercise's current grade (difficulty 3) is probably too low. And it probably shouldn't be unlocked by `hello-world`. That said, I wouldn't worry about this too much, as we'll be re-doing this metadata completely when we launch the v3 track.

As for the actual implementation, I don't mind what is used, as the example implementation is only there to prove that the tests _can_ be solved by some implementation.

> Add some language and a link in the readme pointing the student toward sequences

I don't think we do this for the exercises, where we explicitly point the student towards a specific solution.

jrr:
Thanks Erik.

@MagnusMikkelsen let's keep the implementation and tweak the config. How about `"unlocked_by": "raindrops"` and `"difficulty": 4`?

ErikSchierboom:
Sounds good.

MagnusMikkelsen:
> let's keep the implementation and tweak the config. How about `"unlocked_by": "raindrops"` and `"difficulty": 4`?

Done 👍🏻

jrr:
Merged. Thanks @MagnusMikkelsen !

MagnusMikkelsen:
My pleasure, I had fun. Thanks for all your hard work 👍🏻22. maj 2020 20.53 skrev John Ruble <notifications@github.com>:
Merged. Thanks @MagnusMikkelsen !

PR #818, opened 2020-09-25
[affine-cipher] Correct parameter order of stub functions

ErikSchierboom:
The order did not match the order the tests expect.


robkeim (approved):
Thanks @ErikSchierboom!

PR #832, opened 2020-12-02
fix build

jrr:
The build was failing due to an old `setup_dotnet` GHA task version, which seems to internally use the now-disabled `set-env`:

https://github.com/exercism/fsharp/runs/1480591608?check_suite_focus=true

This PR updates it, and pins a version rather than a commit hash.


ErikSchierboom:
Thanks for fixing this!

> This PR updates it, and pins a version rather than a commit hash.

We pinned to a commit hash for security purposes. It is possible to replace tags with different commits so pinning to a version (tag) is a possible attack vector. That said, this particular GHA is built by GitHub itself, so it is highly unlikely.

ErikSchierboom (approved):

PR #883, opened 2021-01-31
Upgrade to net5.0, update nuget packages and fix solution files

valentin-p:
* Move the solution: Exercise,sln into the exercises/practice folder and rename it Practices.sln
* Create a new solution: Concepts.sln for the exercises/concept exercises.
* Upgrade all the practice exercises to net5.0 instead of deprecated netcoreapp3.0
* Update the nugget packages: FsUnit.xUnit from 4.0.2 to 4.0.4 and Microsoft.NET.Test.Sdk from 16.8.0 to 16.8.3.

Out of scope:
* Fix the Generator code to work with the new folder structure
* Fix the Tests.ps1 script for the new folder structure.

PR #884, opened 2021-01-31
Fix tests for concept records. Bandwagoner

valentin-p:
* Pending on the previous PR #883. #883 needs to be merged first to main.
* Fix tests for the concept exercise records with the name Bandwagoner. Using the existing method `isSameTeam` instead of missing method `isDuplicate`.

Out of scope
* Change module from Stats to Bandwagoner in the Exemplar.fs, it is done in the next PR: #885


ErikSchierboom:
@valentin-p Could you rebase this on the latest `main`?

ErikSchierboom (approved):

PR #886, opened 2021-01-31
Improve example in the practice exercise: triangle 

valentin-p:
Improving the example file for the practice exercise triangle.

The warning raised by the IDE is:
`Incomplete pattern matches on this expression. For example, the value '[_;_;_;_]' may indicate a case not covered by the pattern`


ErikSchierboom (approved):
Great catch!

PR #887, opened 2021-01-31
Fix tests.ps1 to pass the practice tests

valentin-p:
* Pending on the previous PR #888. #888 needs to be merged first to main.
* Pending on the previous PR #884. #884 needs to be merged first to main.
* Update the path /exercises to /exercises/practice so the tests can execute.
* Update the test files to pass the practice exercises the way it was passing before the incorporation of v3.

Out of scope:
* Including the concept exercises in the tests would be done in a later PR, this one is just to rehabilitate the functionality

PR #888, opened 2021-01-31
update test.yml action to run tests with net5.0

valentin-p:
* update test.yml action to run tests with net5.0
* This PR should be merged to main before #887 

I have tested it on my forked repository together with #887: https://github.com/valentin-p/fsharp/runs/1802285485

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,12 +4,12 @@ on: [push, pull_request]
 
 jobs:
   test:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@f90c7b395dac7c5a277c1a6d93d5057c1cddb74e # v2
+      - uses: actions/checkout@v2.3.4
       - uses: actions/setup-dotnet@v1.7.2
         with:
-          dotnet-version: "3.0.100"
+          dotnet-version: "5.0.102"
       - run: pwsh ./test.ps1
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


jrr (approved):

.github/workflows/test.yml
```diff
@@ -4,12 +4,12 @@ on: [push, pull_request]
 
 jobs:
   test:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@f90c7b395dac7c5a277c1a6d93d5057c1cddb74e # v2
+      - uses: actions/checkout@v2.3.4
```
ErikSchierboom, line 9: Could you revert this change? We have a security policy where we fix GitHub actions to SHAs instead of tags.

.github/workflows/test.yml
```diff
@@ -4,12 +4,12 @@ on: [push, pull_request]
 
 jobs:
   test:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
```
ErikSchierboom, line 7: Could you revert this change? It is usually better to use a fixed version than just silently being upgraded. I'm also fine with updating this to `ubuntu-20.04`.

ErikSchierboom (changes requested):

.github/workflows/test.yml
valentin-p: reverted

.github/workflows/test.yml
valentin-p: using 20.04

ErikSchierboom (approved):
Great!

PR #895, opened 2021-02-09
Update status of track

ErikSchierboom:
Closes https://github.com/exercism/fsharp/issues/877

### DIFF
--- a/config.json
+++ b/config.json
@@ -3,9 +3,9 @@
   "slug": "fsharp",
   "active": true,
   "status": {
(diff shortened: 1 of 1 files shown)


valentin-p (approved):
Should we run the tests against the test_runner going forward ?

ErikSchierboom:
> Should we run the tests against the test_runner going forward?

Possibly. That _is_ probably the best way to verify it, although it will be quite a bit slower. I'm still thinking about what to do there :)

PR #900, opened 2021-02-20
Change test.ps1 file to pass all the tests

valentin-p:
- All tests pass
- Support  Example and Examplar
- Support concept and practice exercises
- Pass PSScriptAnalyzer (except one plural)
- Use parameters in functions

To pass both CI checks:
Depends on #901


ErikSchierboom (approved):
Thanks

PR #918, opened 2021-05-14
Pass F# as second item for "Four-item list with fsharp as second item is not exciting"

ceddlyburge:
Presumably this is just a typo ...


ErikSchierboom (approved):

ErikSchierboom:
Thanks!

PR #966, opened 2021-08-27
Copy prerequisites and practices from C#

ErikSchierboom:
@iHiD This is not perfect by any means, but I think it's "Good enough™"


iHiD (approved):
Nice :)

PR #989, opened 2021-09-27
Match function name in documentation to code

onpikono:
Documentation uses function name `incrementDayCount` while the function name defined in BirdWatcher.fs and tested in BirdWatcherTest.fs is `incrementTodaysCount`.


ErikSchierboom (approved):
Great catch!

PR #990, opened 2021-09-28
Update instructions to match tests

onpikono:
Instructions hinted incorrectly on the expected result returned by `description` function. I propose this change to match the tests.


ErikSchierboom (approved):
Excellent! Thanks

PR #991, opened 2021-09-28
Add `open System.Globalization` to be able to use `CultureInfo` class

onpikono:
I propose adding `open System.Globalization` to "hint" students and lead them to explore the `CultureInfo` class instead of manually formatting the date string. Of course, the latter remains as a possibility.

PR #992, opened 2021-09-28
Update instructions with `caution` block

onpikono:
As discussed under PR #991, I have added the `exercism/caution` block to emphasise culture is pre-set in tests.

Closes https://github.com/exercism/fsharp/pull/991


ErikSchierboom (approved):
Great, thanks!

PR #999, opened 2021-10-12
Update of instructions

onpikono:
Hi, I have tried to clean up the instructions to this exercise a bit and possibly make it a bit more informative by:
* Changing the markdown text from plain to code for important values
* Splitting the encoding (encryption) and decoding (decryption) into two sections
* Added guidance about digits and spaces as inputs
* Updated Decoding formula to avoid ambiguity
* Updated the MMI property formula to be more consistent with the resource (Wiki), and subjectively easier to grasp
* Adding a hint about what coprimes (or relative primes) are
* Removed one example of MMI (seems like it's redundant)

Please review and let me know what you think. Thanks!

### DIFF
--- a/exercises/practice/affine-cipher/.docs/instructions.md
+++ b/exercises/practice/affine-cipher/.docs/instructions.md
@@ -1,67 +1,72 @@
-# Instructions
+# Description
 
 Create an implementation of the affine cipher,
 an ancient encryption system created in the Middle East.
 
-The affine cipher is a type of monoalphabetic substitution cipher.
+The affine cipher is a type of mono-alphabetic substitution cipher.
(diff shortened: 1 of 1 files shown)


onpikono:
The commit message says I have "added an example of MMI calculation" which is not the case (I removed that line from the above comment already).

ErikSchierboom:
@onpikono Thanks for this! The instructions are actually copied from the following source file: https://github.com/exercism/problem-specifications/blob/main/exercises/affine-cipher/description.md That file is shared amongst tracks. Could you send a PR to that file instead? That way all tracks can benefit from this improvement. Once merged, we can then sync the F# version from the modified original.

Issue #962, opened 2021-08-20
Check docs are up to date


ErikSchierboom:
Please check if your documentation files are still up-to-date.

The key documentation files to check are:

- `docs/ABOUT.md`
- `docs/INSTALLATION.md`
- `docs/LEARNING.md`
- `docs/RESOURCES.md`
- `docs/TESTS.md`
- `exercises/shared/.docs/help.md`
- `exercises/shared/.docs/tests.md`

There might be more.

## Link check report

To help identify invalid links, we've automatically checked the links of _all_ `*.md` files in this repo.
This is the report of that check:

<pre style="color:#bbb;white-space:pre-wrap;word-wrap:break-word;overflow-wrap:break-word">📝 Summary
---------------------
🔍 Total...........65
✅ Successful......62
⏳ Timeouts.........0
🔀 Redirected.......0
👻 Excluded.........0
🚫 Errors...........3

Errors in docs/ABOUT.md
<span style="color:#ff5555">✗ http://www.tryfsharp.org/Explore (HTTP status client error (404 Not Found) for url (https://redirect.microsoft/Explore))</span>

Errors in docs/LEARNING.md
<span style="color:#ff5555">✗ https://fsharp.org/mentorship/about.html (HTTP status client error (404 Not Found) for url (https://fsharp.org/mentorship/about.html))</span>
<span style="color:#ff5555">✗ https://www.youtube.com/watch?v=7z_q06HQLes (HTTP status client error (403 Forbidden) for url (https://www.youtube.com/oembed?url=https%3A%2F%2Fwww.youtube.com%2Fwatch%3Fv%3D7z_q06HQLes))</span>

</pre>
## Tracking

https://github.com/exercism/v3-launch/issues/54

aage:
@ErikSchierboom 
I've submitted a Pull Request:
https://github.com/exercism/fsharp/pull/1000

aage:
@ErikSchierboom I've reviewed your feedback and updated the Pull Request.

PR #1000, opened 2021-10-12
Update documentation Url's that result in a HTTP 404 NotFound (#962)

aage:


aage:
@ErikSchierboom I've updated all links that pointed to https://github.com/MicrosoftDocs/visualfsharpdocs besides the (two) links for `BigInt`. I couldn't find any content within https://fsharp.github.io/fsharp-core-docs that I could use.

There's a few options (I think):
* Leave it referencing to the markdown file
* Reference to the generic .NET documentation for `BigInteger`: https://docs.microsoft.com/en-us/dotnet/api/system.numerics.biginteger?view=net-5.0
* Reference to another site.

What do you think?

I've again squashed the changes in the initial commit.

Issue #1015, opened 2021-10-27
tracks-on-tracks-on-tracks: add example for discards in list pattern


ErikSchierboom:
The `tracks-on-tracks-on-tracks` exercise wants students to use a discard in their implementation. However, the introduction.md document of that exercise does not explain how that works. We should add a small example on how to use discards in list patterns. Note: this very same change needs to happen in the `introduction.md` document of the `lists` concept.

aage:
I have a question about this exercise in general; is the intent of the exercise _not_ to use the default functions within the `List module` (`List.rev` etc.)?
If not, this isn't made clear in the instruction I think. If yes, then you wouldn't necessarily need to use a discard to solve this exercise, do you?

aage:
@ErikSchierboom I've added a PR with a suggestion for this issues: https://github.com/exercism/fsharp/pull/1028
* You mention that the same change is needed in the `lists` concept, where can I find this? If you like this change I can also add it there.
* In the example _above_ my commited change, the `rec` keyword is used in the `describe` function, but is unnecessary since there's no recursion used. I've omitted this in my change, but would suggest this be removed from that example as well. What do you think?

ErikSchierboom:
@aage Nice! I've commented.

> You mention that the same change is needed in the lists concept, where can I find this? If you like this change I can also add it there.

Each concept has its own `introduction.md` and `about.md` documents. For the `lists` concept, those are found at:

- https://github.com/exercism/fsharp/blob/main/concepts/lists/introduction.md
- https://github.com/exercism/fsharp/blob/main/concepts/lists/about.md

You can basically copy the changes from your PR to those files. In the `about.md` file, you're free to expand on things a bit more if needed (e.g. you can add links).

> In the example above my commited change, the rec keyword is used in the describe function, but is unnecessary since there's no recursion used. I've omitted this in my change, but would suggest this be removed from that example as well. What do you think?

That should absolutely be removed. A separate PR is much appreciated!

aage:
@ErikSchierboom This issue can be closed right? Since the PR is merged into main.

ErikSchierboom:
Correct! Small note: you can use specific keywords to reference an issue from a PR: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword. E.g. if you use `Closes #123`, upon merging the PR issue 123 would be closed.

PR #1028, opened 2021-11-08
Add a discard example when pattern matching on a list for the tracks-…

aage:
…on-tracks-on-tracks exercise #1015

### DIFF
--- a/exercises/concept/tracks-on-tracks-on-tracks/.docs/introduction.md
+++ b/exercises/concept/tracks-on-tracks-on-tracks/.docs/introduction.md
@@ -40,11 +40,13 @@ let describe list =
     match list with
     | [] -> "Empty list"
     | [x] -> "List with one item"
-    | _ -> "List with many items"
+    | [_; y] -> "List with two items (first item ignored)"
+    | _ -> "List with many items (all items ignored)"
 
 describe []        // => "Empty list"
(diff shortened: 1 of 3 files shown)


exercises/concept/tracks-on-tracks-on-tracks/.docs/introduction.md
```diff
@@ -32,3 +32,19 @@ describe []        // => "Empty list"
...
 describe [5; 7; 9] // => "Non-empty with head: 5"
 ```
+
+You can also _discard_ a value when pattern matching; when you do _not_ care about a value in a specific case (i.e. you aren't going to _use_ a value) you can use an underscore (`'_'`) to signify this:
+
+```fsharp
+let describe list =
+    match list with
+    | [] -> "Empty list"
+    | [x] -> "List with one item"
```
ErikSchierboom, line 42: Maybe add an example where one value is ignored?

```suggestion
    | [x] -> "List with one item"
    | [_; y] -> "List with two items (first item ignored)"
```

ErikSchierboom (commented):
One tiny nit.

aage:
@ErikSchierboom I've updated the PR, could you have a look?

ErikSchierboom (approved):
Nice!

PR #1029, opened 2021-11-08
Alphametics tests to handle timeouts for longer tests

martinfreedman:
Changed alphameticstests.fs to make all tests async and added xunit Timeout attribute to longest 3 tests with 2 secs for ``Puzzle with eight letters`` and 5 seconds each for ``Puzzle with ten letters and 199 addends`` and ``Puzzle with ten letters`` exercises.

And switch back on testrunner for this exercise.

### DIFF
--- a/exercises/practice/alphametics/AlphameticsTests.fs
+++ b/exercises/practice/alphametics/AlphameticsTests.fs
@@ -6,7 +6,7 @@ open Xunit
 open Alphametics
 
 [<Fact>]
-let ``Puzzle with three letters`` () =
+let ``Puzzle with three letters`` () = async {
     let puzzle = "I + BB == ILL"
     let expected = 
         [ ('I', 1);
@@ -14,22 +14,25 @@ let ``Puzzle with three letters`` () =
           ('L', 0) ]
         |> Map.ofList
         |> Some
-    solve puzzle |> should equal expected
+    return solve puzzle |> should equal expected
+    }
 
 [<Fact(Skip = "Remove this Skip property to run this test")>]
-let ``Solution must have unique value for each letter`` () =
+let ``Solution must have unique value for each letter`` () = async {
     let puzzle = "A == B"
     let expected = None
-    solve puzzle |> should equal expected
+    return solve puzzle |> should equal expected
+    }
 
 [<Fact(Skip = "Remove this Skip property to run this test")>]
-let ``Leading zero solution is invalid`` () =
+let ``Leading zero solution is invalid`` () = async {
     let puzzle = "ACA + DD == BD"
     let expected = None
-    solve puzzle |> should equal expected
+    return solve puzzle |> should equal expected
+    }
 
 [<Fact(Skip = "Remove this Skip property to run this test")>]
-let ``Puzzle with two digits final carry`` () =
+let ``Puzzle with two digits final carry`` () = async {
     let puzzle = "A + A + A + A + A + A + A + A + A + A + A + B == BCC"
     let expected = 
         [ ('A', 9);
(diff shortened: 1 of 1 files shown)


exercises/practice/alphametics/AlphameticsTests.fs
```diff
@@ -78,10 +84,11 @@ let ``Puzzle with seven letters`` () =
           ('T', 7) ]
         |> Map.ofList
         |> Some
-    solve puzzle |> should equal expected
+    return solve puzzle |> should equal expected
+    }
 
-[<Fact(Skip = "Remove this Skip property to run this test")>]
-let ``Puzzle with eight letters`` () =
+[<Fact(Timeout=2000,Skip = "Remove this Skip property to run this test")>]
```
ErikSchierboom, line 90: Nice idea. The actual timeouts are slightly tricky to determine, but I'm fine with having some value here. That said, we currently can't merge this as the test runner will strip the attribute before it runs the tests:

```
type EnableAllTests() =
    inherit SyntaxVisitor()

    override _.VisitSynAttribute(attr: SynAttribute) : SynAttribute =
        match attr.TypeName with
        | LongIdentWithDots ([ ident ], _) when ident.idText = "Fact" ->
            base.VisitSynAttribute
                { attr with
                      ArgExpr = SynExpr.Const(SynConst.Unit, attr.ArgExpr.Range) }
        | _ -> base.VisitSynAttribute(attr)
```

Would you be interested in fixing that too?

martinfreedman: Yup will have a look at that shortly.

martinfreedman: Will look at this. Which repo and file is it?

martinfreedman: I think timeouts should be discussed at the practise exercise specification, hence my other PR to update canonical json with a timeout property. This might include relative timings between tests as implied here between 8 versus 10 letter tests. The actual implementation would depend on the track maintainers factoring in the built in online test runner timeouts and language performance etc. I would probably default to 10 seconds per test, _a propos_ of anything more specific.

ErikSchierboom:
> Which repo and file is it?

This one: https://github.com/exercism/fsharp-test-runner/blob/main/src/Exercism.TestRunner.FSharp/Rewrite.fs#L20

> I think timeouts should be discussed at the practise exercise specification, hence my other PR to update canonical json with a timeout property. This might include relative timings between tests as implied here between 8 versus 10 letter tests. The actual implementation would depend on the track maintainers factoring in the built in online test runner timeouts and language performance etc. I would probably default to 10 seconds per test, a propos of anything more specific.

I'm not entirely sure about the timeouts being in the prob-specs definition, but I'll post my thoughts there.

Oh and I completely forgot to mention that the `alphametics` exercise's test file is automatically generated by a test generator: https://github.com/exercism/fsharp/blob/main/generators/Generators.fs#L61 That means that any changes that are not made to the generator would be overwritten when we re-generate the tests. I think you can ignore that for now, as the F# test generator hasn't yet been updated to use the `tests.toml` file (I started working on that), but I did want to mention it.

ErikSchierboom:
Speaking of the `tests.toml` file, that would be a great place to configure these timeouts.

exercises/practice/alphametics/AlphameticsTests.fs
martinfreedman: I already wrote a reply but it has inexplicably disappeared. What files do I need to look at?

ErikSchierboom: See https://github.com/exercism/fsharp/pull/1029#issuecomment-964015126. I should have posted that comment here.

ErikSchierboom: @martinfreedman Let me know if you need any help/pointers. The F# syntax tree can be tricky to work with.

martinfreedman: Will do. And yes the Diamond exercise does not use `[<Fact>]`! I will add this into my solution when I have a chance (I have Covid right now). Any other exercises tests with non-standard Attributes?

ErikSchierboom: > (I have Covid right now)

Oh no! All the best.

> Any other exercises tests with non-standard Attributes?

The `simple-linked-list` exercise uses a `[<Theory>]` attribute (which should actually be converted to multiple `[<Fact>]` attributes.

PR #1030, opened 2021-11-08
Create example.fs

martinfreedman:
An example file that will pass all tests including timeouts in [PR1029](https://github.com/exercism/fsharp/pull/1029) alphameticstests.fs


martinfreedman:
There is an existing example.fs file that pass the tests including timeouts.

PR #1077, opened 2022-04-27
Clarifying getting values from tuple example

Ymirke:
Hey, I think this example would be a bit more clear if length2 was named length1 in the first example of getting values out of tuples.


ErikSchierboom (approved):
Agreed!

PR #1089, opened 2022-08-10
Fix: Invalid format of interest rates

sinancsoysal:
Correct conversion from percentage should be like:
3.213% -> 0.03213 not 3.213
0.5% -> 0.005 not 0.5
1.621%  -> 0.01621 not 1.621
2.475% -> 0.02475 not 2.475
and so on..

This way users won't have to recalculate correct form of interest rates in every function.


ErikSchierboom:
Thanks for wanting to contribute. Unfortunately, we can't accept this contribution, as it goes against what we're trying to teach the student. Having the interest be specified as it is, we force the student to divide it by 100, which then helps teach the student about division and converting between different numeric types.

PR #1090, opened 2022-08-18
Change example from list to array

timyhac:
This page is for arrays, but this example shows a list.
Further, list elements are not mutable using that assignment operator, so the example will not compile.

### DIFF
--- a/concepts/arrays/about.md
+++ b/concepts/arrays/about.md
@@ -15,7 +15,7 @@ let threeValues = [| "a"; "b"; "c" |]
 Elements can be assigned to an array or retrieved from it [using an index][array-indexer]. F# arrays are zero-based, meaning that the first element's index is always zero:
 
 ```fsharp
-let numbers = [2; 3; 5]
+let numbers = [|2; 3; 5|]
 
 // Update value in array
 numbers.[2] <- 9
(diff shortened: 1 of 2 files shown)


concepts/arrays/about.md
```diff
@@ -71,4 +71,4 @@ describe [5; 7]     // => "Array with two items (first item ignored)"
 describe [5; 7; 9]  // => "Array with many items (all items ignored)"
```
ErikSchierboom, line 71: This section also contains some lists. Could you also update those?

ErikSchierboom (commented):
Good catch! Just one minor comment.

concepts/arrays/about.md
timyhac: Yep, done!

ErikSchierboom (approved):
Sorry about this, but currently a lot of this information is duplicated in https://github.com/exercism/fsharp/blob/main/exercises/concept/bird-watcher/.docs/introduction.md. Could you also update that file? We'll work on removing the duplication later.

ErikSchierboom (approved):

ErikSchierboom:
Merged! Thanks 🎉

PR #1092, opened 2022-09-02
Add representer docs

ErikSchierboom:
- Add representer normalizations doc
- Add representer mentoring doc


docs/representer/MENTORING.md
```diff
@@ -0,0 +1,13 @@
+# Mentoring
```
ErikSchierboom, line 1: We could also consider this document to be a general, track-specific document on mentoring which then could also be used on the general mentoring page. Thoughts?

iHiD (approved):

PR #1139, opened 2023-02-05
[Hamming approaches]: links fix 

meatball133:
Matthijs noticed that the links were broken on the approach, link: https://forum.exercism.org/t/markdown-in-approach-blurb-is-not-rendered/3459

I think it has to do with how the spacing with the links is set up. This pr is meant to fix that.


ErikSchierboom (approved):
Great!

MatthijsBlom:
That forum link is not correct; I noticed it in #1131.

PR #1144, opened 2023-02-07
Arrays concept docs

glennj:
- don't use dot-bracket syntax
- fix some erroneous list syntax in concept exercise introduction

### DIFF
--- a/concepts/arrays/about.md
+++ b/concepts/arrays/about.md
@@ -18,10 +18,10 @@ Elements can be assigned to an array or retrieved from it [using an index][array
 let numbers = [| 2; 3; 5 |]
 
 // Update value in array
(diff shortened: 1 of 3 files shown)


glennj:
Is there a particular reason why the concept exercises don't use templates for their introductions?

ErikSchierboom (approved):
Thanks!

> Is there a particular reason why the concept exercises don't use templates for their introductions?

Yes. They weren't available when I built the exercises and I never took the time to convert them 🤷 Would be useful.

glennj:
> I never took the time to convert them 🤷 Would be useful.

I can follow up

ErikSchierboom:
That would be great! (if you're keen on more work, C# could use the same PR 😉 )

PR #1158, opened 2023-02-25
Clarify Bird Watcher exercise

scubbo:
It is ambiguous to:
* Use the word "_odd_" to describe both index-parity and "_the property of being unusual_".
* Determine parity with one-indexed counting, especially when the introduction to Arrays points out that Array-indexing is zero-based.

I do recognize that both of these issues could be resolved by a student closely reading the examples, and that learning to do due diligence in understanding a task is a core Software Engineer skill. However, Exercism's [mission](https://exercism.org/about) seems to only discuss "_programming_" (not "_engineering_"); which I personally understand as being _purely_ the practice of coding, without the undoubtedly important other career "_soft skills_". That is - that exercises are not intended to be deliberately ambiguous. This is tenuously and tangentially supported by the (excellent) condemnation of ambiguous terms [here](https://exercism.org/docs/community/being-a-good-community-member/the-words-that-we-use).

(I chose to clarify what "_odd_" means, rather than to update the problem definition, because I wasn't sure whether that would invalidate previous solutions)

### DIFF
--- a/exercises/concept/bird-watcher/.docs/hints.md
+++ b/exercises/concept/bird-watcher/.docs/hints.md
@@ -28,7 +28,7 @@
   - The `Array` module has a [function to update a value in an array][set-function].
 - The above function to update a value in an array does not return a value. You should thus manually return the updated array after setting its value.
 
-## 6. Check for odd week
+## 6. Check for unusual week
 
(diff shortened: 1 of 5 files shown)


scubbo:
Thank you! :D

PR #1204, opened 2024-01-06
Add new test for grade-school exersice

Grenkin1988:
Hello,

Exercise have this description:

> Get a sorted list of all students in all grades. Grades should sort as 1, 2, 3, etc., and students within a grade should be sorted alphabetically by name.
"Who all is enrolled in school right now?"
"Let me think. We have Anna, Barb, and Charlie in grade 1, Alex, Peter, and Zoe in grade 2 and Jim in grade 5. So the answer is: Anna, Barb, Charlie, Alex, Peter, Zoe and Jim"

Given that all tests are using `add` function as part of the `school` preparation, we never test that `roster` function handles the case when input `school` was not sorted.

### DIFF
--- a/exercises/practice/grade-school/GradeSchoolTests.fs
+++ b/exercises/practice/grade-school/GradeSchoolTests.fs
@@ -2,76 +2,79 @@ module GradeSchoolTests
 
 open FsUnit.Xunit
 open Xunit
+
 open GradeSchool
 
 [<Fact>]
-let ``Roster is empty when no student is added``() =
+let ``Roster is empty when no student is added`` () =
     let school = empty
     roster school |> should be Empty
 
 [<Fact(Skip = "Remove this Skip property to run this test")>]
-let ``Student is added to the roster``() =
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
Thanks for the PR, but I'm not sure we should be merging this. The problem with the newly added test is that it forces an implementation upon the student. We deliberately don't force the student to use a map, as they're then free to implement as they see fit.

Grenkin1988:
> Thanks for the PR, but I'm not sure we should be merging this. The problem with the newly added test is that it forces an implementation upon the student. We deliberately don't force the student to use a map, as they're then free to implement as they see fit.

Sorry I do not 100% understand this. Do you mean that it is OK for student to remove type alias and do something else instead?

ErikSchierboom:
Oh the type alias is still there? I thought I had removed that 😕 It's fine then!

ErikSchierboom (approved):

Grenkin1988:
Yes, it uses `type School = Map<int, string list>`

PR #1205, opened 2024-01-15
Run `configlet sync --docs` to sync things and update Leap

IsaacG:
+cc @ErikSchierboom


config.json
ErikSchierboom, line 1: Thanks for the format!

ErikSchierboom (approved):
Thanks a lot!

PR #1230, opened 2024-02-08
Update Isogram.fs with type signatures

jurecuhalev:
Adds input and output types to make it clearer what's expected. It also helps IDEs out of the box.


kotp:
https://forum.exercism.org/t/update-isogram-fs-with-type-signatures/9706 for the forum discussion that matches this PR.

Issue #868, opened 2021-01-28
[v3] Implement new Concept Exercise: chars


ErikSchierboom:
This issue describes how to implement the `chars` concept exercise for the F# track.

## Getting started

**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read up on the following documents:

- [The features of v3](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md).
- [Rationale for v3](https://github.com/exercism/v3/blob/master/docs/rationale-for-v3.md).
- [What are concept exercise and how they are structured?](https://github.com/exercism/v3/blob/master/docs/features-of-v3.md)

Please also watch the following video:

- [The Anatomy of a Concept Exercise](https://www.youtube.com/watch?v=gkbBqd7hPrA).

## Goal

The goal of this exercise is to teach the student the basics of the Concept of Chars in F#. We'll also use this exercise to introduce `for..in` loops and a mutable binding. 

## Learning objectives

- Know of the existence of the `char` type.
- Know what a char represents (a Unicode character).
- Know that chars are immutable.
- Know how to define a char.
- Know how to access a char in a string by index.
- Know of some basic char methods (like converting to uppercase).
- Know how to convert an integer to a character and vice versa.

## Out of scope

- Converting a string to a char array.

## Concepts

The Concepts this exercise unlocks are:

- `chars`: know of the existence of the `char` type; know what a char represents (a Unicode character); know that chars are immutable; know how to define a char; know how to access a char in a string by index; know of some basic char methods (like converting to uppercase); know how to convert an integer to a character and vice versa.

## Prerequisites

This exercise's prerequisites Concepts are:

- `strings`: know about working with strings.

### Hints

- [Char methods][char-methods]: lists the available methods on the `Char` class.

Once a reference document has been written on the subject of chars in F# (see section 11), it should be added too.

### After

- [Char methods][char-methods]: lists the available methods on the `Char` class.

Once a reference document has been written on the subject of chars in F# (see section 11), it should be added too.

## Representer

This exercise does not require any specific representation logic to be added to the [representer][representer].

## Analyzer

This exercise does not require any specific logic to be added to the [analyzer][analyzer].

## Implementing

To implement this exercise, please [follow these instructions](https://github.com/exercism/v3/blob/master/languages/fsharp/reference/implementing-a-concept-exercise.md).

## Help

[...]

PR #1248, opened 2024-04-18
[Squeaky Clean] new char concept and concept exercise

colinleach:
[As discussed](https://forum.exercism.org/t/f-learning-syllabus/10835/3), this PR attempts to follow Issue #868.

It's mostly a port from C#. I had to drop the task using `Char.IsControl`, as F# rejects things like `\0` (nul) as characters.

The `about.md` is much expanded, with lots more sample code. More controversially, so is `introduction.md` - now too much, perhaps?

### DIFF
--- a/concepts/chars/.meta/config.json
+++ b/concepts/chars/.meta/config.json
@@ -1,5 +1,4 @@
 {
   "blurb": "A char is a type that represents a Unicode UTF-16 character. Strings are a sequence of chars.",
-  "authors": ["colinleach"],
-  "contributors": ["mikedamay"]
+  "authors": ["colinleach"]
 }
--- a/concepts/chars/.meta/config.json
+++ b/concepts/chars/.meta/config.json
@@ -1,0 +1,5 @@
+{
+  "blurb": "A char is a type that represents a Unicode UTF-16 character. Strings are a sequence of chars.",
+  "authors": ["colinleach"],
+  "contributors": ["mikedamay"]
+}
--- a/concepts/chars/introduction.md
+++ b/concepts/chars/introduction.md
@@ -2,10 +2,87 @@
 
 The F# `char` type is a 16 bit quantity to represent the smallest addressable components of text.
 
-Multiple `char`s can comprise a string such as `"word"` or `char`s can be processed independently. 
+Multiple `char`s can comprise a string such as `"word"` or `char`s can be processed independently.
 
-Their literals have single quotes e.g. `'A'`.
+They can be defined as literals with single quotes:
 
-F# `char`s support UTF-16 Unicode encoding so in addition to the latin character set pretty much all the writing systems used worldwide can be represented, e.g. the Greek letter `'β'`.
+```fsharp
+let ch = 'A'
+=> val ch: char = 'A'
+```
 
-There are many builtin library methods to inspect and manipulate `char`s. These can be found as static methods of the `System.Char` class.
+An individual `char` can be retrieved from a string with (zero-based) indexing:
+
+```fsharp
+Exercism"[4]
+=> val it: char = 'c'
+```
+
+Iterating over a string returns a `char` at each step:
+
+```fsharp
+[| for c in "F#" -> c, int c |]
+=> val it: (char * int) array = [|('F', 70); ('#', 35)|]
+```
+
+As shown above, a `char` can be cast to its `int` value.
+This also works (*at least some of the time*) for other scripts:
+
+```fsharp
+[| for c in "東京" -> c, int c |] // Tokyo, if Wikipedia is to be believed
+=> val it: (char * int) array = [|('東', 26481); ('京', 20140)|]
+```
+
+The underlying Int16 is used when comparing characters:
+
+```fsharp
+'A' < 'D'
+=> val it: bool = true
+```
+
+Also, an `int` can be cast to `char`:
+
+```fsharp
+char 77
+=> val it: char = 'M'
+```
+
+The `System.Char` library contains the full set of [methods][Char-methods] expected for a .NET language, such as upper/lower conversions (but see the caveats in the next section):
+
+```fsharp
+'a' |> System.Char.ToUpper
+=> val it: char = 'A'
+
+'Q' |> System.Char.ToLower
+=> val it: char = 'q'
+```
+
+To append a `char` to a `string`, convert it to a string first:
+
+```fsharp
+"Part" + string '1'
+=> val it: string = "Part1"
+```
+
+Using a StringBuilder may be more performant:
+
+```fsharp
+let sb = new System.Text.StringBuilder()
+=> val sb: System.Text.StringBuilder = 
+
+sb.Append('F').Append('#')
+=> val it: System.Text.StringBuilder = F# {Capacity = 16; Chars = ?; Length = 2; MaxCapacity = 2147483647;}
+
+sb.ToString()
+=> val it: string = "F#"
+```
+
+Converting between a `list` of `char` and a `string` is simple:
+
+```fsharp
+System.String.Concat ['E'; 'x'; 'e'; 'r'; 'c'; 'i'; 's'; 'm']
+=> val it: string = "Exercism"
+
+"Exercism" |> Seq.toList
+=> val it: char list = ['E'; 'x'; 'e'; 'r'; 'c'; 'i'; 's'; 'm']
+```
--- a/concepts/chars/introduction.md
+++ b/concepts/chars/introduction.md
@@ -45,7 +45,8 @@ Also, an `int` can be cast to `char`:
 char 77  // => 'M'
 ```
 
-The `System.Char` library contains the full set of methods expected for a .NET language, such as upper/lower conversions:
+The `System.Char` class has many useful methods, such as upper/lower conversions:
+
 
 ```fsharp
 'a' |> System.Char.ToUpper  // =>  'A'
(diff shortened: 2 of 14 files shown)


colinleach:
Some CI fails at the first attempt.

It dislikes the example code in my md files, in this format:

```fsharp
'A' < 'D'
=> val it: bool = true
```
I'll comment out the FSI output and resubmit.

I don't know why it thinks the `.meta/config.json` is "unformatted". Locally, `configlet lint` was happy with it, and I can't see a problem (yet).

colinleach:
Docs are now fixed and passing. I still has an unexplained problem with `exercises/concept/squeaky-clean/.meta/config.json`

Also, I suspect the prereqs will need some attention. It only lists `strings`, which may not be enough. I relied on lists, the C# version relies on mutable variables (loop counter).

concepts/chars/introduction.md
```diff
@@ -0,0 +1,88 @@
...
+
+Converting between a `list` of `char` and a `string` is simple:
+
+```fsharp
+System.String.Concat ['E'; 'x'; 'e'; 'r'; 'c'; 'i'; 's'; 'm']
+// => val it: string = "Exercism"
+
+"Exercism" |> Seq.toList
+// => val it: char list = ['E'; 'x'; 'e'; 'r'; 'c'; 'i'; 's'; 'm']
+```
```
ErikSchierboom, line 88: This introduces lists, which we would need to have as a prerequisite (might be fine).

ErikSchierboom (commented):
Thanks for this! I think the key thing is to get the exemplar right. I feel like it might be a tid too complex for this exercise, so we might need to tweak the exercise a bit.

ErikSchierboom:
I've played around a bit with the exercise, and what if we did something like this:

```fsharp
module SqueakyClean

open System

let transform (c: char) : string =    
    if c = '-' then
        "_"
    elif c > 'α' && c < 'ω'  then
        "?"
    elif Char.IsWhiteSpace(c) then
        ""
    elif Char.IsUpper(c) then
        "-{c.ToLower()}
    else
        c.ToString()

let clean (identifier: string): string =
    String.collect transform identifier 
```

This seems to check all the boxes I'd like to see ticked:

- Use character literals (`'-'`, `'α'` and `'ω'`)
- Use static `System.Char` methods (like `IsWhiteSpace()`)
- Use `System.Char` methods (like `ToUpper()`)
- Compare `char` instances via `=`, `<` and `>`
- Convert from `char` to `string`
- "Append" a char to a string via string interpolation (or `sprintf` for now, until we've updated the `log-levels` exercise)
- Show that strings are a sequence of characters (albeit a bit indirectly)

The benefits of this approach are:

- We don't have to get in mutable territory, which a for loop would force us into
- We don't need StringBuilder
- We don't need pattern matching, but can instead just use a regular if-else (we need the `"if-then-else-expressions"` prerequiste)
- We have students use the `String` module

The downside is:

- We don't have student's use a `for` loop, but instead a higher-order function is used (we should do some handwaving here, as I don't feel like higher-order functions should be a depedency)

What do you think?

concepts/chars/.meta/config.json
```diff
@@ -0,0 +1,5 @@
+{
+  "blurb": "A char is a type that represents a Unicode UTF-16 character. Strings are a sequence of chars.",
+  "authors": ["colinleach"],
+  "contributors": ["mikedamay"]
```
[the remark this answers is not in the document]
colinleach: OK. My thinking was that the original draft was mostly his text copied verbatim, but the docs have evolved since then.

colinleach:
Yes! Your exercise suggestion is much better. I'll need to rewrite the instructions and tests (and take StringBuilder out of the blurb), but that should be easy. I'm 9 h behind you, so give me a day or so,

> I don't feel like higher-order functions should be a dependency

For many reasons, including the small detail that we don't yet have an exercise for this, and the concept is a very short stub.

concepts/chars/introduction.md
```diff
@@ -0,0 +1,88 @@
...
+'Q' |> System.Char.ToLower
+// => val it: char = 'q'
+```
+
+To append a `char` to a `string`, convert it to a string first:
+
+```fsharp
+"Part" + string '1'
+// => val it: string = "Part1"
+```
```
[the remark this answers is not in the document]
colinleach: > To append a `char` to a `string`, convert it to a string first

As you maybe guessed, this was a half-assed last-minute addition that I sensibly left out of earlier drafts. Trying/failing to make the exercise work early in the syllabus when the student doesn't know much.

concepts/chars/introduction.md
colinleach: I'm increasingly thinking that lists would be a good prereq, moving `chars` down the syllabus.

colinleach:
Overall, the message I'm taking is that I should pay less attention to the C# syllabus and treat F# as a very different thing. I can't disagree.

ErikSchierboom:
> Overall, the message I'm taking is that I should pay less attention to the C# syllabus and treat F# as a very different thing. I can't disagree.

It might work well in some cases! But probably a combination of C# and Elixir would be good. But I'm also happy to review like this. It wasn't much work for me!

colinleach:
- Exercise is reworked based (pretty closely) on your proposal.
- Both `introduction.md` files are shortened.
- Added `if-then-else-expressions` as a prereq.

I added a link to the `higher-order-functions` concept `about.md`, so that could usefully be expanded.

With string interpolation in there, that needs adding to `log-levels` as you point out.

> a higher-order function is used (we should do some handwaving here, as I don't feel like higher-order functions should be a dependency)

I added  a draft of this handwaving, but I suspect it could be improved.

exercises/concept/squeaky-clean/SqueakyClean.fs
```diff
@@ -0,0 +1,6 @@
+module SqueakyClean
+
+open System
```
ErikSchierboom, line 3: Good idea to already have this opened!

concepts/chars/introduction.md
```diff
@@ -0,0 +1,88 @@
+# Introduction
+
+The F# `char` type is a 16 bit quantity to represent the smallest addressable components of text.
+
+Multiple `char`s can comprise a string such as `"word"` or `char`s can be processed independently.
```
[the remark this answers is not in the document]
ErikSchierboom: I see you resolved this commend but I still see the old text?

concepts/chars/introduction.md
```diff
@@ -0,0 +1,53 @@
...
+```fsharp
+[| for c in "F#" -> c, int c |]  //  => [|('F', 70); ('#', 35)|]
+```
+
+As shown above, a `char` can be cast to its `int` value.
+This also works (*at least some of the time*) for other scripts:
+
+```fsharp
+[| for c in "東京" -> c, int c |] //  =>  [|('東', 26481); ('京', 20140)|]
+```
```
ErikSchierboom, line 33: Maybe instead of having a list comprehension as the example here, we use a higher-order function?

concepts/chars/introduction.md
```diff
@@ -0,0 +1,53 @@
...
+
+The underlying Int16 is used when comparing characters:
```
ErikSchierboom, line 35:
```suggestion
The underlying `Int16` type is used when comparing characters:

```

concepts/chars/introduction.md
```diff
@@ -0,0 +1,53 @@
...
+'A' < 'D'  // =>  true
+```
+
+Also, an `int` can be cast to `char`:
+
+```fsharp
+char 77  // => 'M'
+```
+
+The `System.Char` library contains the full set of methods expected for a .NET language, such as upper/lower conversions:
```
ErikSchierboom, line 47: `System.Char` is not really a library:

```suggestion
The `System.Char` class has many useful methods, such as upper/lower conversions:

```

exercises/concept/squeaky-clean/SqueakyCleanTests.fs
```diff
@@ -0,0 +1,47 @@
...
+
+open FsUnit.Xunit
+open Xunit
+open Exercism.Tests
+
+open SqueakyClean
+
+[<Fact>]
+[<Task(1)>]
+let ``Clean_empty_string``() = clean "" |> should equal ""
```
ErikSchierboom, line 11: I think that it would be useful for the student to have them implement the `let transform (c: char) : string` function first, before having to deal with the higher-order function. So we'd have the tests work on single characters first, and then later on have a couple of tests that verify that `clean` can work on an entire string.

This also has the benefit that it is more likely that students use the same exemplar solution, which gives great for mentoring.

ErikSchierboom (commented):
Great work!

concepts/chars/introduction.md
colinleach: Agree. Update will follow later today.

exercises/concept/squeaky-clean/SqueakyCleanTests.fs
colinleach: Agree. Update will follow later today.

concepts/chars/introduction.md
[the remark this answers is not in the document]
colinleach: I may have hit the wrong button (I had a few health problems over the weekend). Apologies.

colinleach:
I hope these changes address your comments (and I didn't carelessly miss any this time). I also added a couple of minor tweaks to try and make the docs flow more coherently.

ErikSchierboom (approved):
I think this is excellent.I'm second-guessing myself a little bit on whether we want to do the full string processing at all, but it's probably fine. We can always move that bit out of this exercise and in a new exercise, for example one that has one deal with StringBuilder

PR #1268, opened 2024-06-08
Update instructions for lucians-luscious-lasagna with datatype

s-praveen-kumar:
This concept introduces datatypes, both type inference and explicit type annotation is covered. As a beginner I felt really confused as to what type should I annotate my functions with and is it even needed. (I was trying uint and was constantly getting lots of errors). Hence adding a note on the type expectation for future learners.

### DIFF
--- a/exercises/concept/lucians-luscious-lasagna/.docs/instructions.md
+++ b/exercises/concept/lucians-luscious-lasagna/.docs/instructions.md
@@ -2,7 +2,7 @@
 
 In this exercise you're going to write some code to help you cook a brilliant lasagna from your favorite cooking book.
 
-You have four tasks, all related to the time spent cooking the lasagna.
+You have four tasks, all related to the time spent cooking the lasagna. Use `int` as type for inputs and output.
 
 ## 1. Define the expected oven time in minutes
 
--- a/exercises/concept/lucians-luscious-lasagna/.docs/instructions.md
+++ b/exercises/concept/lucians-luscious-lasagna/.docs/instructions.md
@@ -2,7 +2,12 @@
 
 In this exercise you're going to write some code to help you cook a brilliant lasagna from your favorite cooking book.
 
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
The introduction has this section:

> Bindings are defined using the let keyword. Specifying a binding's type is optional for most bindings, as F#'s type inference can usually infer the type based on their value. A binding looks like this:

Your argument is that this is not enough to go on?

s-praveen-kumar:
Hi @ErikSchierboom 
My point was, the learning material introduces both type inference and explicit type annotation. But I was not sure which one to use for the exercise. 

At least when I attempted as a beginner, I was annotating different types and was getting error messages.

ErikSchierboom:
Makes sense. I don't have much time now, but I'll come back to this!

exercises/concept/lucians-luscious-lasagna/.docs/instructions.md
```diff
@@ -2,7 +2,7 @@
 
 In this exercise you're going to write some code to help you cook a brilliant lasagna from your favorite cooking book.
 
-You have four tasks, all related to the time spent cooking the lasagna.
+You have four tasks, all related to the time spent cooking the lasagna. Use `int` as type for inputs and output.
```
ErikSchierboom, line 5:
```suggestion
You have four tasks, all related to the time spent cooking the lasagna.

~~~~exercism/note
You can either use type inference, where you don't specify any types, or use explicit typing, where you do specify the types.
If you choose the latter, use `int` as the type.
~~~~
```

ErikSchierboom: @s-praveen-kumar What do you think about the above suggestion?

s-praveen-kumar: Yes @ErikSchierboom . This is much clearer

ErikSchierboom (approved):
👍

PR #1308, opened 2025-03-01
rest-api: replace triple-quote strings with single-quote ones to mitigate the issue with the test runner

RamanBut-Husaim:
## Description

The issue is described in details on the forum https://forum.exercism.org/t/restapi-and-system-text-json-issue/15946 and inside the [issue](https://github.com/exercism/fsharp-test-runner/issues/272) with the `fsharp-test-runner`.

The existing rewrite approach in `fsharp-test-runner` does not escape properly triple-quoted strings with a backslash `\` symbol inside. After the rewrite it produces an invalid json inside a string by double-escaping the `"` symbol.

## Solution

A quick fix that I verified locally on my machine is to replace triple-quoted strings with a single-quoted one. The rewrite operation works fine with such cases. The quick fix was agreed with @ErikSchierboom inside the issue on the forum.

The proper fix is to do a more thorough investigation of the `rewrite` operation in the test runner and address the root cause.


ErikSchierboom:
Thanks!

PR #1323, opened 2025-08-01
Fix StateOfTicTacToe tests

Sprootch:
Typo in gamestate instead of gameState.


ErikSchierboom:
Normally I'd not merge this due to it breaking the existing solutions, but as there are so few, it's okay in this case. You do also need to update this file: https://github.com/Sprootch/exercism-fsharp-pr/blob/18517e8423365cbf37b3bd97b41ca9ec78a76aed/exercises/practice/state-of-tic-tac-toe/.meta/Example.fs#L26

PR #1376, opened 2026-03-21
Validate Exercise.sln exercise list in test.ps1

blackk-foxx:
This will fail CI if the exercise sets in `config.json` and `exercises/Exercises.sln` differ, as discussed in http://forum.exercism.org/t/working-solution-to-passwordchecker-failing-on-exercism-infrastructure/30559/17.

### DIFF
--- a/bin/test.ps1
+++ b/bin/test.ps1
@@ -190,7 +190,7 @@ function Assert-Exercise-List-Agreement {
     $config_filename = "config.json"
     $sln_filename = "exercises/Exercises.slnx"
 
-    if ($concept_config_sln_diff || $concept_sln_config_diff || $practice_config_sln_diff || $practice_sln_config_diff) {
+    if ($concept_config_sln_diff -or $concept_sln_config_diff -or $practice_config_sln_diff -or $practice_sln_config_diff) {
         Write-Output "Error: exercise sets in $config_filename and $sln_filename differ."
         Report-Diff "Concept" $config_filename $sln_filename $concept_config_sln_diff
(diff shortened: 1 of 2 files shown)


ErikSchierboom (commented):
Thanks! I see one problem with the current solution: it requires several tools that are not natively installed on the systems of maintainers (especially not on Windows). Maybe you could have this solution work with just the functionality that is in PowerShell/.NET? I know for example that there is JSON support in PowerShell. It will result in more code, but it'll be cross-platform and not require anyone to install additional software.

blackk-foxx:
> The failure behavior is fine. I don't mind that. The solution path is wrong though, it's an `.slnx` file

OK

blackk-foxx:
> It looks good. I can verify that if I omit an exercise from the solution, the diff is shown. However, if I remove an exercise from the config.json but still have it as a project, it does _not_ work. That is actually totally fine, because `configlet lint` will warn about this. I would remove those checks such that you only end up with comparing the config.json's exercise to the solution.

It turned out that I was using the wrong OR operator.  This was causing practice exercises discrepancies to be ignored, in addition to the concept exercises in config.json but not in Exercises.sln.  Now that it is fixed, it will catch any discrepancy between config.json and Exercises.sln, both for concept and practice exercises.  For completeness,I would prefer to keep the in-config.json-but-not-in-Exercises.sln checks, but if you feel strongly about it, I will remove them.

ErikSchierboom:
Works great. Thanks!

PR #1384, opened 2026-06-27
Add `save-the-cow`, deprecating `hangman`

BNAndras:
Related to https://forum.exercism.org/t/what-frp-library-is-needed-for-fsharp-hangman/61381

### DIFF
--- a/config.json
+++ b/config.json
@@ -1950,7 +1950,7 @@
           "records",
           "sets"
         ],
-        "difficulty": 7
+        "difficulty": 8
       },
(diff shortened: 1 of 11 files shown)


BNAndras:
The example solution and test suite are brought over verbatim from Hangman. There were some minor differences between the local Hangman test suite and problem specs version, but I kept those as-is since this exercise is a drop-in replacement for Hangman.

ErikSchierboom:
Thanks!

